PR #17, opened 2015-03-02
Switch to use local bootstrap repo instead of collecting from many locations

tfoote:
This will increase reliability and consistency. And make things more reproducable. We just have to maintain that one repo.


import_upstream.bash
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+
+export PYTHONPATH=/home/rosbuild/reprepro_updater/src
```
wjwwood: Should we be hard coding this? This is a public repository, I think it shouldn't contain the folder structure of our server if possible.

tfoote: This repo originally was just a place for me to collect helper scripts. 

This master branch is only used for our legacy system. All scripts have the full paths hard coded and will never be released. There's the refactor branch which is where things have been reworked for potential release, and get their configuration from external parameterization.

esteve:
I had a look at this and the other two PR, and the changes make sense to me. Is there a way to test this?

tfoote:
I have tested https://github.com/ros-infrastructure/buildfarm_deployment_config/pull/4 in my docker based environment but the others are modifications of the live system.

import_upstream.bash
dirk-thomas: The `refactor` branch also contains absolute paths. They should be replaced but not necessarily part of this PR.

dirk-thomas:
LGTM

esteve:
+1

PR #18, opened 2015-03-08
add saucy for ceres import

mikeferguson:
This was not originally building correctly when the import was added, but it does now exist in the PPA (and has for some time). I just finally released the first thing that depends on libceres-dev into Indigo, and noticed that it failed to build on Saucy because of this.


tfoote:
@mikeferguson FYI this database is now contained here: https://github.com/ros-infrastructure/ros_bootstrap_dependencies/blob/master/ubuntu/config/ceres.upstream.yaml

mikeferguson:
@tfoote -- thanks, wasn't aware of the new location!

PR #19, opened 2015-09-09
Add importer for sixad 1.6+ from mikepurvis.

mikepurvis:
Saucy is EOL, so Launchpad won't let me push a release for that. As such, I've left it pulling qtsixa, and added sixad 1.6+ as pulling for trusty and vivid (utopic is also EOL).

More rationale here: https://github.com/ros/rosdistro/pull/9297#issuecomment-138983648

@mikeferguson, @tfoote


mikepurvis:
Ping.

mikeferguson:
+1 (we actually sync'd this into our mirror already so we are using it on all of our robots)

PR #20, opened 2015-11-06
compute package diffs in in sync jobs, to be logged in console outputs

tfoote:
connects to ros-infrastructure/ros_buildfarm#79


tfoote:
Fixes https://github.com/ros-infrastructure/ros_buildfarm/issues/79

I've extended the pr to include the output into the console output of the sync. Demonstrated in manually reconfigured http://54.183.26.131:8080/view/Manage/job/Jrel_sync-packages-to-testing_trusty_amd64/291/consoleFull#console-section-8

tfoote:
@dirk-thomas please review

dirk-thomas:
The code is not valid Python 3 code and you might want to run a code style checker.

scripts/diff_packages.py
```diff
@@ -1,144 +1,56 @@
...
+        args.sourcerepo,
+        args.dist,
+        args.component,
+        args.arch
+    )
+
+    try:
+        pf_old = diff_repos.get_packagefile_from_url(target_url, name='old')
+    except RuntimeError, ex:
+        parser.error("%s" % ex)
```
dirk-thomas: Instead of `"%s" % ex`: `str(ex)`.

Same below.

scripts/diff_packages.py
```diff
@@ -1,144 +1,56 @@
...
-        fh.write("\n\n")
 
-        fh.write("Thanks to all ROS maintainers who make packages"
-                 " available to the ROS community. The above list "
-                 "of packages was made possible by the work of the"
-                 " following maintainers:\n")
-        for maintainer in sorted(maintainers):
-            fh.write(" * %s\n" % maintainer)
+    print("Difference between %s and %s computed at %s" %
+          (target_url, upstream_url, dtime.isoformat('-')))
```
dirk-thomas: Such placeholders should be quoted for better readability.

Same below.

scripts/sync_ros_packages.py
```diff
@@ -90,6 +90,37 @@
...
+            package_architecture)
+        upstream_url = diff_repos.construct_packages_url(
+            upstream_repo_url,
+            ubuntu_distro,
+            'main',
+            package_architecture)
+        try:
+            pf_old = diff_repos.get_packagefile_from_url(target_url, name='old')
+        except RuntimeError, ex:
+            print("Exeception: %s \n NOT Computing diff" % ex, file=sys.stderr)
```
dirk-thomas: Are these exceptions not fatal? They will for sure not be noted if the job is still "green".

src/reprepro_updater/diff_repos.py
```diff
@@ -0,0 +1,131 @@
...
+    maintainers = set()
+    for p in added_packages | updated_packages:
+        maintainers.add(strip_email(new_packages[p]['Maintainer']).strip())
+
+    class OutputAppender:
+        def __init__(self):
+            self.value = ""
+
+        def append(self, string_value):
+            self.value += string_value + '\n'
```
dirk-thomas: Why a custom class for this? A simple list would do the job the same way.

src/reprepro_updater/diff_repos.py
```diff
@@ -0,0 +1,131 @@
...
+    out.append("Removed Packages [%s]:\n" % len(removed_packages))
+    for p in sorted(removed_packages):
+        out.append("- %s" % (p))
+    out.append("\n")
+
+    out.append(
+        "Thanks to all ROS maintainers who make packages"
+        " available to the ROS community. The above list "
+        "of packages was made possible by the work of the"
+        " following maintainers:")
```
dirk-thomas: Alternating (leading / trailing) white spaces?

dirk-thomas:
I don't know enough how the code in this repo works. So I only commented on code stuff.

scripts/sync_ros_packages.py
tfoote: Failures are not fatal. It doesn't work correctly at the moment for source changes(it needs to find and unzip the Source.gz instead of the straight download) . And it will be noticed by anyone who is going to write an announcement.

tfoote:
Resolved comments, Made it python3 compatible and added some basic unit test for the new functionality.

dirk-thomas:
Still pep8 E226 two times.

tfoote:
Fixed up for pep8. Do we have a canonical pep8 configuration file? My instance had E226 off by default which is the pep8 default. [default configuration](https://pep8.readthedocs.org/en/latest/intro.html#configuration)

dirk-thomas:
https://github.com/ament/ament_lint/blob/master/ament_pep8/ament_pep8/configuration/ament_pep8.ini

tfoote:
Thanks, I added a link to the ROS2 developer guide.

PR #24, opened 2016-02-25
add rule to import backport of trollius to trusty

wjwwood:
This is needed to release `osrf_pycommon` on Trusty. The `python[3]-trollius` debs start up in vivid, so we'll have trusty and vivid+. Since I am using launchpad, I can't easily create backports for EOL Ubuntu distributions (like utopic).

### DIFF
--- a/config/python-trollius.backport.yaml
+++ b/config/python-trollius.backport.yaml
@@ -1,0 +1,6 @@
+name: trollius
+method: http://ppa.launchpad.net/wjwwood/trollius-backports/ubuntu
+suites: [trusty]
+component: main
+architectures: [any]
+filter_formula: Package (== python-trollius) | Package (== python3-trollius)


config/python-trollius.backport.yaml
```diff
@@ -0,0 +1,6 @@
+name: trollius
+method: http://ppa.launchpad.net/wjwwood/trollius-backports/ubuntu
+suites: [trusty]
+component: main
+architectures: [any]
```
tfoote: apparently in this context neither `any` nor `all` work in this context. I think we're going to have to enumerate the supported arches unfortunately.

tfoote:
Manually merged here: ce49e9ff1c4e99e60aa04279e01152da7952bd3a and fixedup here: c0e6c56b64b08d22375a334118f21d0e81f5bc8f

wjwwood:
Thanks.

wjwwood:
@tfoote I can see `python-trollius` in repositories.ros.org, but not packages.ros.org yet. Is there a job on the farm that should be run to sync it to OSU or should I do it the old way? Can I do it or are we waiting to do so based on something else?

tfoote:
@wjwwood same process. i forgot to sync to osu. They are there now.

PR #25, opened 2016-03-10
add wily to suites to import for collada-dom

wjwwood:
See: https://github.com/ros/rosdistro/issues/9945#issuecomment-194642340


wjwwood:
We may have to include Xenial in this list but with a partial import rule (not sure that's possible), see: https://github.com/ros/rosdistro/pull/10673

tfoote:
lgtm

k-okada:
@wjwwood I'm not sure how this works, but can we put arm deb to ros repository if we add `armhf`?

tfoote:
We can also import armhf and arm64 debs if you build them.

PR #26, opened 2016-03-11
added arm debs

k-okada:
is this ok? we have trusty, vivid, wily, we do not have arm deb for saucy and utopic due to they are EOL on launchpad builder.


tfoote:
yeah, that should import them from the same repos.

tfoote:
imported to buildfarm, they will go out with the next syncs

PR #37, opened 2017-03-06
add catkin-lint yakkety

mikaelarguedas:
Looks like there are no builds for zesty [on the ppa](http://ppa.launchpad.net/roehling/latest/ubuntu/pool/main/c/catkin-lint/) at the moment so I'm adding only yakkety. Is this the way to go or should I add zesty too and it will just start working once the deb is available ?

### DIFF
--- a/config/catkin_lint.upstream.yaml
+++ b/config/catkin_lint.upstream.yaml
@@ -1,6 +1,6 @@
 name: catkinlint
 method: http://ppa.launchpad.net/roehling/latest/ubuntu
-suites: [precise, trusty, utopic, vivid, wily, xenial, yakkety]
+suites: [precise, trusty, utopic, vivid, wily, xenial, yakkety, zesty]
 component: main
 architectures: [i386, amd64, source]
 filter_formula: Package (== catkin-lint ) | Package (% python*-catkin-lint)
--- a/config/catkin_lint.upstream.yaml
+++ b/config/catkin_lint.upstream.yaml
@@ -1,6 +1,6 @@
 name: catkinlint
 method: http://ppa.launchpad.net/roehling/latest/ubuntu
-suites: [precise, trusty, utopic, vivid, wily, xenial]
+suites: [precise, trusty, utopic, vivid, wily, xenial, yakkety]
 component: main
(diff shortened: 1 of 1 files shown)


tfoote:
@roehling any chance you could turn on zesty builds?

Also have you considered submitting this upstream for packaging? We could make an introduction to maintainers if that would be helpful.

roehling:
Done.

You mean submitting to Ubuntu? I haven't thought of it yet, but I like the idea.

tfoote:
@roehling Thanks for the new version. Yeah, I was thinking upstream to Ubuntu and actually Debian. Ubuntu pulls their packages by default from debian periodically for each release cycle. The great thing is that once it's packaged it will generally appear on most of our common platforms for each release cycle.

@j-rivero could you give @roehling some pointers into the process of submitting for upstream packaging?

@mikaelarguedas FYI Zesty is now available too

j-rivero:
> @j-rivero could you give @roehling some pointers into the process of submitting for upstream packaging?

Sure. The [official introduction](https://wiki.debian.org/Packaging/Intro?action=show&redirect=IntroDebianPackaging) is good for getting the basic packaging concepts, although I prefer to work with git repositories (I mean use git for the debian packaging tasks nothing related to the package official source) and for this the gbp tool is great. There is a [good tutorial](http://honk.sigxcpu.org/projects/git-buildpackage/manual-html/gbp.html) about how to use it with the workflow explained.

Debian packaging is quite picky about the policies required. The [Policy Manual](https://www.debian.org/doc/debian-policy/) is extensive but is the best source in this topic. Since this catkin-lint looks like a python package, the [python documentation for packaging](https://wiki.debian.org/Python/LibraryStyleGuide) is specially interesting. You can use `apt-get source <package>` to explore how the different Debian packages implement their packaging data (look inside the `debian/` directory)

Once we have the package ready, we will need to upload it to debian-science and ask for a sponsorship. I can take care of this part easily.

mikaelarguedas:
thanks @roehling for the new version. PR updated accordingly

tfoote:
@mikaelarguedas imported

PR #38, opened 2017-03-10
Use source package in update rule filter

roehling:
By using the `Source` filter, the update rule will continue to function even if binary packages are renamed or removed. Everything built from the `catkin-lint` source package will be matched by the rule.


tfoote:
Great thanks!

PR #42, opened 2017-04-25
adding pyside2 source entry

### LINKED ISSUE #39: Add import of pyside2 source debs
Requested here: https://discourse.ros.org/t/debian-source-package-for-pyside2/1572/3

tfoote:
Fixes #39

I tested the import rule and there's a new version upstream. Here's the dry run of the update: https://gist.github.com/tfoote/3227e3a70406067e77159072b508ce75

@dirk-thomas Running this as is will bring in a new build. Do we want to do so?


dirk-thomas:
A newer version is fine with me :+1:

tfoote:
imported: http://build.ros.org:8080/job/import_upstream/118/

PR #40, opened 2017-04-18
Gazebo8 rules

mikaelarguedas:
This adds rules to import gazebo8 for Lunar.

Needed before https://github.com/ros/rosdistro/pull/14564 can be merged

PR #43, opened 2017-04-27
Missing gazebo8 packages

j-rivero:
I forgot `gazebo8-common` in https://github.com/ros-infrastructure/reprepro-updater/pull/40 and the -dbg packages is `gazebo8-dbg` instead of `libgazebo8-dbg`


tfoote:
Imported to ROS repos here: http://build.ros.org:8080/job/import_upstream/120/console

PR #49, opened 2017-05-24
Fail script in case of failed subcommands

dirk-thomas:
Based on top of #48.

If any of the invoked subcommands failed (e.g. search for "returned non-zero exit status" in http://build.ros.org:8080/job/import_upstream/125/consoleFull) the script should fail and not "hide" the problem in the hundres of kb of output.

### DIFF
--- a/src/reprepro_updater/helpers.py
+++ b/src/reprepro_updater/helpers.py
@@ -162,8 +162,10 @@ def run_cleanup(repo_dir, rosdistro, distro, arch, commit):
     lockfile = os.path.join(repo_dir, 'lock')
     with LockContext(lockfile) as lock_c:
 
-        _clear_ros_distro(repo_dir, rosdistro, distro, arch, commit)
-        delete_unreferenced(repo_dir, commit)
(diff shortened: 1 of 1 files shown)


src/reprepro_updater/helpers.py
```diff
@@ -157,8 +162,10 @@ def run_cleanup(repo_dir, rosdistro, distro, arch, commit):
     lockfile = os.path.join(repo_dir, 'lock')
     with LockContext(lockfile) as lock_c:
 
-        _clear_ros_distro(repo_dir, rosdistro, distro, arch, commit)
-        delete_unreferenced(repo_dir, commit)
+        if not _clear_ros_distro(repo_dir, rosdistro, distro, arch, commit):
+            raise RuntimeError('command failed')
```
tfoote: These look good however we should have unique descriptive outputs for the three different failures.

dirk-thomas: Done: f98181cdc558f0f9fa30c24bb9753a23cf02e2dc

PR #56, opened 2018-02-26
Improve performance of include_folder.py

nuclearsandwich:
I spent time Friday and Saturday trying to investigate the performance issues ticketed in https://github.com/ros-infrastructure/buildfarm_deployment/issues/165

### Test case

My main test case is rebuilding catkin on indigo trusty amd64

My test farm isn't quite the same as the production farm. It's a little under-powered but that makes it easier to test the impact of performance improvements.

With a fresh import of indigo from the production repositories the first catkin import package job takes ~2.5 hours. Which is enough to make the catkin build job timeout and require a second run to complete successfully. Since the import package job does clean out the repositories, subsequent package imports complete within normal range because everything must be rebuilt.

### Optimization strategy

Since we had fewer issues with the same reprepro version (5.1.1) on the Trusty farm, which was using the zesty amd64 deb copied directly rather than a homemade backport of the zesty deb for Xenial, which is what the current farm uses, I first tried installing that zesty deb on my testfarm. This resulted in no change to job time or behavior.

reprepro's apparent maintenance status, our [other issues with it](https://github.com/ros-infrastructure/buildfarm_deployment/issues/186), and the community momentum behind [aptly](https://www.aptly.info/) make us reluctant to invest heavily in trying to improve reprepro itself so for the time being we're going to try and optimize our usage of reprepro.

### First pass

The first optimization was to remove all transitive dependencies with a single invocation of reprepro removefilter. Each run of removefilter required a reindexing to occur.  While I haven't profiled the reindexing separate from the rest of the remove operation. Running the remove filter takes 2.0-2.6s so doing that once instead of ~2000 times will definitely pay dividends.

### Next steps

The next major performance blocker is fetching the transitive dependencies. The current method uses a listfilter query and parses the results, adding each new dependent package to a queue to be processed. This is roughly equivalent to the recursive function tracking seen packages that was in use previously. Both rely on one exec of reprepro listfilter per package to process. Which means that finding catkin's transitive deps requires something like 2000 execs, each of which takes ~1.6sec.

To resolve this it's occurred to me so far to

#### throw compute at it to try and reduce the exec time

This is just straight throwing money at the problem :money_with_wings: I'm not even sure it can help that much since it we'll always have the overhead of the individual execs.

#### parallelize dependency collection

This will also require throwing some extra money at compute in order to get a machine with more processors/threads. We would be able to run a couple execs at once but this wouldn't reduce the overall number of execs and it'd mean making sure that this part of the code is concurrency-safe.

#### Find a way to get all transitive dependents at once

As near as I can tell, there's no way to get transitive dependents with a single reprepro filterlist. Which means that we would need to use something that isn't reprepro.
This is kind of the magic ideal as it replaces a 2k iterative/recursive process with a single-step one. I have no idea how to do it and I am not convinced that it's possible without precompiling the list of transitive dependents for each package.

#### Bring dependent fetching logic in-process

Instead of invoking reprepro filterlist to determine which packages depend on a given package read the repository manifest and do so ourselves in-process. Since I haven't profiled reprepro itself (and ideally don't intend to) I don't know how much time it spends starting up then reading and parsing repository information but it probably isn't negligible. If we can reduce the iteration time even a little we'll reduce the overall runtime significantly.

### DIFF
--- a/src/reprepro_updater/helpers.py
+++ b/src/reprepro_updater/helpers.py
@@ -167,10 +167,10 @@ def invalidate_dependent(repo_dir, distro, arch, package):
     # on the buildfarm when invalidating low-level packages.
     repo_info = RepositoryInfo(repo_dir, distro, arch)
 
-    # queue of dependents to iterate
-    dependents_to_process = repo_info.get_rdepends(package)
-    # the complete list of transitive dependents
-    transitive_dependents = dependents_to_process.copy()
+    # We'll build this into a list of transitive dependents.
+    transitive_dependents = repo_info.get_rdepends(package)
+    # queue of dependents to iterate, seeded with the initial rdepends.
+    dependents_to_process = list(transitive_dependents)
 
     queue_max_size = len(dependents_to_process)
     processed_packages = 0
@@ -179,7 +179,7 @@ def invalidate_dependent(repo_dir, distro, arch, package):
         _start = time.time()
         dep = dependents_to_process.pop()
         depdeps = repo_info.get_rdepends(dep)
-        dependents_to_process |= (depdeps - transitive_dependents)
+        dependents_to_process += (depdeps - transitive_dependents)
         transitive_dependents |= depdeps
 
         if len(dependents_to_process) > queue_max_size:
--- a/src/reprepro_updater/helpers.py
+++ b/src/reprepro_updater/helpers.py
@@ -129,23 +129,11 @@ def invalidate_dependent(repo_dir, distro, arch, package):
     # queue of dependents to iterate, seeded with the initial rdepends.
     dependents_to_process = list(transitive_dependents)
 
-    queue_max_size = len(dependents_to_process)
-    processed_packages = 0
-    iter_times = []
     while dependents_to_process:
-        _start = time.time()
         dep = dependents_to_process.pop()
         depdeps = repo_info.get_rdepends(dep)
         dependents_to_process += (depdeps - transitive_dependents)
         transitive_dependents |= depdeps
-
-        if len(dependents_to_process) > queue_max_size:
-            queue_max_size = len(dependents_to_process)
-        processed_packages += 1
-        iter_times.append(time.time() - _start)
-        print("Queue size:\t\t{}".format(len(dependents_to_process)))
-        print("Dependents found:\t{}".format(len(transitive_dependents)))
-        print("Avg iter time:\t\t{}".format(sum(iter_times) / len(iter_times)))
 
     return invalidate_packages(repo_dir, distro, arch, transitive_dependents)
 
--- a/src/reprepro_updater/helpers.py
+++ b/src/reprepro_updater/helpers.py
@@ -80,53 +80,6 @@ def _run_update_command(repo_dir, distro, commit):
     return try_run_command(update_command)
 
 
-def _get_dependent_packages(repo_dir, distro, arch, package):
-    """
-    Return a list of packages which were detected as dependent by reprepro.
-
-    This only returns direct dependencies.
-    """
-    reprepro_command = [
-        'reprepro', '-V', '-b', repo_dir,
-        '-T', 'deb',
-        'listfilter', distro,
-        "Package (% ros-* ), " +
-        "Architecture (== " + arch + " ), " +
-        "( Depends (% *" + package + "[, ]* ) " +
-        "| Depends (% *" + package + " ) )"]
-
-    try:
-        output = subprocess.check_output(reprepro_command)
-    except subprocess.CalledProcessError as ex:
-        print("Execution of [%s] Failed:" % reprepro_command, ex)
-        return []
-    packages = []
-    for l in output.splitlines():
-        elements = l.split()
-        assert len(elements) == 3, "Expected 3 elements, got %s " % elements
-        packages.append(elements[1])
-    return packages
-
-
-def _invalidate_dependent(repo_dir, distro, arch, package, processed_packages):
-    """Implement invalidation for recursion."""
-    # Get all dependents and recursively walk their dependents
-    dependents = _get_dependent_packages(repo_dir, distro, arch, package)
-    for dependent in dependents:
-        # packages may overlap in the walk at different depths
-        # don't try to redelete a package if it's already been processed
-        if dependent in processed_packages:
-            continue
-        # walk to all dependencies
-        if not _invalidate_dependent(repo_dir, distro, arch, dependent, processed_packages):
-            return False
-        # clear the package
-        if not invalidate_package(repo_dir, distro, arch, dependent):
-            return False
-        processed_packages.append(dependent)
-    return True
-
-
 def invalidate_packages(repo_dir, distro, arch, packages):
     """
     Remove multiple packages from the repo in one reprepro invocation.
--- a/src/reprepro_updater/helpers.py
+++ b/src/reprepro_updater/helpers.py
@@ -159,9 +159,17 @@ def invalidate_dependent(repo_dir, distro, arch, package):
 
     This is only valid for binary packages.
     """
-    # storage for recursion
-    processed_packages = []
-    return _invalidate_dependent(repo_dir, distro, arch, package, processed_packages)
+    # queue of dependents to iterate
+    dependents_to_process = set(_get_dependent_packages(repo_dir, distro, arch, package))
+    # the complete list of transitive dependents
+    transitive_dependents = dependents_to_process.copy()
+    while len(dependents_to_process) > 0:
+        dep = dependents_to_process.pop
+        depdeps = set(_get_dependent_packages(repo_dir, distro, arch, dep))
+        dependents_to_process |= (depdeps - transitive_dependents)
+        transitive_dependents |= depdeps
+
+    return invalidate_packages(repo_dir, distro, arch, transitive_dependents)
 
 
 def _clear_ros_distro(repo_dir, rosdistro, distro, arch, commit):
--- a/src/reprepro_updater/helpers.py
+++ b/src/reprepro_updater/helpers.py
@@ -164,7 +164,7 @@ def invalidate_dependent(repo_dir, distro, arch, package):
     # the complete list of transitive dependents
     transitive_dependents = dependents_to_process.copy()
     while len(dependents_to_process) > 0:
-        dep = dependents_to_process.pop
+        dep = dependents_to_process.pop()
         depdeps = set(_get_dependent_packages(repo_dir, distro, arch, dep))
         dependents_to_process |= (depdeps - transitive_dependents)
         transitive_dependents |= depdeps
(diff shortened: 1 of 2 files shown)


nuclearsandwich:
While writing up the PR my latest test build at the current state finished. It cuts the job time from 2.5hours to 1.2 (so just about in half).

As I said previously, I don't think this is enough to mitigate the issue but it's a fine start.

tfoote:
A few other options: only use the ROS dependencies for invalidation. Then catkin_pkg can provide the reverse lookup. 

*In-process*
I think that parsing the Packages file wouldn't be too hard in python. I think there might be even an importer that can parse it, possibly as ini or as json. You might need to split it on blank lines or the like first. That would allow you to get all the transitive dependencies in one shot.

nuclearsandwich:
I did spend time yesterday and this morning bringing the parsing of the Packages file to seek rdepends in-process. The result is a resounding success!

I haven't found a way to reduce the runtime order of finding transitive dependencies, but what was killing us anyway was the high constant cost of each iteration. I'm happy to say that reading the Packages file in python and parsing out package dependencies once before repeatedly iterating to find transitive dependendents is indeed _way_ faster than repeatedly invoking reprepro.

The average time of one iteration fell from 1.6sec to 0.0007sec bringing the total testfarm time to import catkin on indigo from 2.5hrs (1.2hr after initial pass) to 12 seconds.

The three tested configurations: the current refactor branch, this pr at 
876dcb3, and this pr at 7f58002 all removed the same 2271 packages (albeit in a different order but I don't think that's significant).

My next testing phase will be to make sure this isn't a fluke by adding kinetic to my testfarm and importing catkin once with the default reprepro and then once with this branch.

nuclearsandwich:
While I set up additional test cases please review the python and docstrings. We could move the RepositoryInfo class into the helpers file but it is getting a bit crowded in there.

src/reprepro_updater/helpers.py
```diff
@@ -142,11 +159,38 @@ def invalidate_dependent(repo_dir, distro, arch, package):
...
+
+    # queue of dependents to iterate
+    dependents_to_process = repo_info.get_rdepends(package)
+    # the complete list of transitive dependents
+    transitive_dependents = dependents_to_process.copy()
+
+    queue_max_size = len(dependents_to_process)
+    processed_packages = 0
+    iter_times = []
+    while len(dependents_to_process) > 0:
```
dirk-thomas: Shorter: `while dependents_to_process`

src/reprepro_updater/helpers.py
```diff
@@ -142,11 +159,38 @@ def invalidate_dependent(repo_dir, distro, arch, package):
...
+        _start = time.time()
+        dep = dependents_to_process.pop()
```
dirk-thomas: This processes the pending items in a random order. In case something goes wrong it will be almost impossible to reproduce the problem.

src/reprepro_updater/helpers.py
```diff
@@ -142,11 +159,38 @@ def invalidate_dependent(repo_dir, distro, arch, package):
...
+    # on the buildfarm when invalidating low-level packages.
+    repo_info = RepositoryInfo(repo_dir, distro, arch)
+
+    # queue of dependents to iterate
+    dependents_to_process = repo_info.get_rdepends(package)
+    # the complete list of transitive dependents
+    transitive_dependents = dependents_to_process.copy()
+
+    queue_max_size = len(dependents_to_process)
+    processed_packages = 0
```
dirk-thomas: These two variables don't seem to be used anywhere?

src/reprepro_updater/repository_info.py
```diff
@@ -0,0 +1,72 @@
...
+        the dependencies into memory.
+
+        Sets the variable package_dependencies which is a dict of package names
+        mapping to sets of dependency names.
+        """
+        if self.package_dependencies is None:
+            self.package_dependencies = {}
+        packages_filepath = os.path.join(self.repo_dir, 'dists', self.distro,
+                                         'main', 'binary-{}'.format(self.arch), 'Packages')
+        packages_contents = open(packages_filepath, "r").read()
```
dirk-thomas: Use `with` to ensure the file is properly closed.

src/reprepro_updater/repository_info.py
```diff
@@ -0,0 +1,72 @@
...
+                if line.startswith("Depends: "):
+                    depends = line.split(": ")[1]
+            if name is None:
+                raise RuntimeError(
+                    "Repository file '{}' had a section missing 'Package':\n+\n{}\n+".format(
+                        packages_filepath, section))
+            if depends is None:
+                dependency_set = set()
+            else:
+                dependency_set = set([item.strip().split(' ')[0] for item in depends.split(',')])
```
dirk-thomas: I would suggest to perform this set comprehension right after the value is extracted from the file (line 38). By initializing `depends` in the beginning you can then get rid of the second variable `dependency_set`.

src/reprepro_updater/helpers.py
```diff
@@ -142,11 +159,38 @@ def invalidate_dependent(repo_dir, distro, arch, package):
...
-    # storage for recursion
-    processed_packages = []
-    return _invalidate_dependent(repo_dir, distro, arch, package, processed_packages)
+    # Use internal repository information parser to get reverse dependents.
+    # Using reprepro for each package was resulting in huge runtimes causing timeouts
+    # on the buildfarm when invalidating low-level packages.
+    repo_info = RepositoryInfo(repo_dir, distro, arch)
+
+    # queue of dependents to iterate
+    dependents_to_process = repo_info.get_rdepends(package)
```
dirk-thomas: Since this line doesn't copy the returned set it will mutate the data later on (calling `pop()`) which is likely not desired.

src/reprepro_updater/repository_info.py
```diff
@@ -0,0 +1,69 @@
...
+                    name = line.split(': ')[1]
+                if line.startswith('Depends: '):
+                    depends = {
+                        item.strip().split(' ')[0] for item in line.split(': ')[1].split(',')}
+            if name is None:
+                raise RuntimeError(
+                    "Repository file '{}' had a section missing 'Package':\n+\n{}\n+".format(
+                        packages_filepath, section))
+            if depends is None:
+                depends = set()
```
dirk-thomas: You can avoid these two lines by initializing `depends` with an empty set in line 31 already.

src/reprepro_updater/helpers.py
nuclearsandwich: Do you have a suggestion about what to do differently? Using a list instead would make every unique addition linear on the size of the queue.

While I expect the set ordering to be arbitrary I would expect it to be deterministic on constant inputs so this does not concern me much.

src/reprepro_updater/helpers.py
nuclearsandwich: This is a good point and is true of anything that gets a set from get_rdepends. I think It's best resolved by copying the memoized set when returning it rather than making each caller remember to do so.

nuclearsandwich: The returned set is now a copy of the internal one as of 9839c51

src/reprepro_updater/helpers.py
dirk-thomas: > Do you have a suggestion about what to do differently? Using a list instead would make every unique addition linear on the size of the queue.

You would need a list, yes. Every `append` is certainly not `O(n)` but only `O(1)` (see https://wiki.python.org/moin/TimeComplexity). Of course the sorting will take some time but I don't expect it to be significant and it would pay of in case of issues.

> While I expect the set ordering to be arbitrary I would expect it to be deterministic on constant inputs so this does not concern me much.

I don't think so. That would make it deterministic for the same input which it is not.

nuclearsandwich: > Every append is certainly not O(n) but only O(1) (see https://wiki.python.org/moin/TimeComplexity

Yes, but if we naively append every new element we'll process packages that appear multiple times in the dependency tree... multiple times. 

> That would make it deterministic for the same input which it is not.

You're right. It's deterministic within an ipython context but not in general.

src/reprepro_updater/repository_info.py
```diff
@@ -0,0 +1,72 @@
...
+class RepositoryInfo:
+    """
+    Parse apt repository data in order to provide information for reprepro-updater.
+    """
+    def __init__(self, repo_dir, distro, arch):
+        self.repo_dir = repo_dir
+        self.distro = distro
+        self.arch = arch
+        self.package_dependencies = None
+        self.package_rdepends = None
```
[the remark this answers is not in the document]
tfoote: It would probably need to invoke the `_parse_package_file` here, not just the empty initializer. But still it would simplify the logic.

nuclearsandwich:
I noticed a cross-rosdistro dependency issue when testing kinetic (which is xenial based like lunar) and found out that my behavior didn't match the original behavior. 
285dcde addresses that and https://github.com/j-rivero/fcl0.5/issues/1 is opened to address the cross-distro dependency.

nuclearsandwich:
Testing invalidation of ros-kinetic-catkin is now compatible with the original behavior. Importing jessie packages to test on a debian distro. Then I think I'm going to build a significant chunk of kinetic on xenial in order to test that there aren't any nasty surprises as we build further up the rosdistro.

I don't think it makes sense to try and build up a comprehensive test case building an entire rosdistro with both the existing and improved branches and compare each package since it would require writing up a significant amount of test machinery to pull and compare jenkins logs each step of the way. But if we wanted to do so I could start working on that.

nuclearsandwich:
Alright, I'm resetting my test environment to get some final tests:

1. Testing that this code works for the jessie repositories in Kinetic.
2. Building kinetic for a couple of hours without import failures then trying to install some built package from the test repository in a docker container and verifying that it can be installed. I'm not going to perform any tests on the built package, I just want to make sure a package and all its dependencies are properly part of the repository.

nuclearsandwich:
After testing I'll remove a bunch of the perf output and get the PR ready for a final review.

@tfoote is there a sufficient benefit to a percentage-based or scoped rollout to these changes so we can monitor further without the possibility of a complete disruption?  To do either would require adding additional code here and/or in ros_buildfarm to handle it. Some possibilities.

1. Deploy as is so all buildfarms targeting the refactor branch always use the new implementation.
2. Add a switch that routes invalidations to the new implementation or old based on a random switch. i.e. start at 20% and scale up to 100% through a followup branch.
3. Use a switch and restrict this change to indigo/kinetic for a time?

I don't see much reason to delay rollout if testing is positive, I just wanted to present it as an option.

nuclearsandwich:
Found and fixed a failure when no dependent packages are found.

nuclearsandwich:
> This looks pretty good. It looks like there might be some vestigial code to clean up?

Yeah. I have been leaving code in for now in the event that we'd want to stagger rollout of the new method. This has been building kinetic on the testfarm for the 15 hours. If it hits 24 I'll call it good, strip out the reporting and vestigial code, then after a final review pass we can merge.

nuclearsandwich:
The latest changes (dead code and timing output removed) has been building on a testfarm overnight with no complaints. I believe this is ready to deploy.

PR #58, opened 2018-04-20
Add mirroring of gazebo9 for artful and stretch.

clalancette:
Signed-off-by: Chris Lalancette <clalancette@openrobotics.org>

@j-rivero , a quick look over this would be appreciated.  @tfoote , once we are happy with this and it is merged I'll ping you for actually getting it deployed.


j-rivero:
> @j-rivero , a quick look over this would be appreciated

I miss the `-dev` versions of the ignition packages (transport4, msgs, etc). They are dependencies of `libgazebo9-dev`.

clalancette:
> I miss the -dev versions of the ignition packages (transport4, msgs, etc). They are dependencies of libgazebo9-dev.

Excellent point.  I fixed this in 2c5ab4d.  Thanks for the review.

config/gazebo9.upstream.yaml
```diff
@@ -0,0 +1,6 @@
+name: gazebo9_import
```
mikaelarguedas: should this name and the file have the platform name as well ? (like you did for stretch below)

config/gazebo9_debian_stretch.upstream.yaml
```diff
@@ -0,0 +1,6 @@
+name: gazebo9_import_debian_stretch
+method: http://packages.osrfoundation.org/gazebo/debian-stable
+suites: [stretch]
+component: main
+architectures: [amd64, source]
```
mikaelarguedas: We also need arm64 right?

config/gazebo9.upstream.yaml
clalancette: I don't have a strong feeling either way.  The existing configs have a bunch of styles.  I guess more information is better, so I'll change it (and the name of the file) to be more clear.

config/gazebo9_debian_stretch.upstream.yaml
clalancette: I had checked on artful, and there weren't arm64 packages available, but I forgot to double check on stretch.  Good call, I'll add arm64 here.

mikaelarguedas: As we don't build debs for arm on artful that's not a big deal. For debian we do need the arm64 ones at least.
Then it would be convenient for users if we this would allow people to build from source, but not a requirement

config/gazebo9_debian_stretch.upstream.yaml
```diff
@@ -0,0 +1,6 @@
+name: gazebo9_import_debian_stretch
+method: http://packages.osrfoundation.org/gazebo/debian-stable
+suites: [stretch]
+component: main
+architectures: [amd64, arm64, armhf, source]
```
mikaelarguedas: if we add armhf here, should we also add i386 ?

tfoote: That's a good question. I'd say we should bring in all arches that are available. Here and on Artful. It will help anyone who wants to build from source.

j-rivero:
> Excellent point. I fixed this in 2c5ab4d. Thanks for the review.

No problem. Looks good to me now.

config/gazebo9_debian_stretch.upstream.yaml
mikaelarguedas: I see this PR has been merged, shoud this be addressed in a follow up then ?

tfoote: Yeah I was just about to comment back. I figured I'd get this merged and we could come back for i386 if there was demand for it.

tfoote:
imported: http://build.ros.org/job/import_upstream/201/

clalancette:
Excellent, thanks!

PR #59, opened 2018-05-08
Include ignition-cmake in the gazebo9 sync for Debian Strech and Ubuntu Artful

j-rivero:
I forgot to include ignition-cmake in the sync for gazebo9 into the ROS repositories. It is a dependency of ignition-msgs. //cc @clalancette


mikaelarguedas:
Not sure if it matters but stretch has 0.4.1 while artful and bionic has 0.6.0.
Are both versions ok ?

clalancette:
> Not sure if it matters but stretch has 0.4.1 while artful and bionic has 0.6.0.
> Are both versions ok ?

Actually, it is going to be a little stranger than that.  For bionic, we are getting the `libignition-cmake-dev` package from upstream Ubuntu, where it is 0.4.0.  For artful, we'll be getting the latest version of `libignition-cmake-dev` from the mirror of the gazebo repositories (http://packages.osrfoundation.org/gazebo/ubuntu-stable/pool/main/i/ignition-cmake/ , so 0.6.0 like you point out).  For stretch, we'll be getting the latest version of `libignition-cmake-dev` from the mirror of the gazebo repositories (http://packages.osrfoundation.org/gazebo/debian-stable/pool/main/i/ignition-cmake/, so 0.4.1).  It doesn't change your fundamental question, but it is a bit odd.  @j-rivero , is that all going to be OK?

j-rivero:
> For bionic, we are getting the libignition-cmake-dev package from upstream Ubuntu, where it is 0.4.0. For artful, we'll be getting the latest version of libignition-cmake-dev from the mirror of the gazebo repositories (http://packages.osrfoundation.org/gazebo/ubuntu-stable/pool/main/i/ignition-cmake/ , so 0.6.0 like you point out). For stretch, we'll be getting the latest version of libignition-cmake-dev from the mirror of the gazebo repositories (http://packages.osrfoundation.org/gazebo/debian-stable/pool/main/i/ignition-cmake/, so 0.4.1). It doesn't change your fundamental question, but it is a bit odd. @j-rivero , is that all going to be OK?

Well it is hard to say that 'all is going to be ok' having different versions available in different platforms. We are support the default installation from Ubuntu (using 0.4.0) that should be fine with gazebo9. I can release 0.6.0 into Stretch to have same version that in Artful.

//cc @scpeters @mxgrey: can you spot any problem [between 0.4.0 and 0.6.0](https://bitbucket.org/ignitionrobotics/ign-cmake/compare/ignition-cmake_0.6.0%0Dignition-cmake_0.4.0#commits) versions of ign-cmake for using with gazebo9?

mxgrey:
The changes between 0.4.0 and 0.6.0 shouldn't have a noticeable impact on Gazebo. The most observable difference for Gazebo is that in 0.6.0 we're marking some cache variables as advanced; in 0.4.0, those cache variables may clutter a user's cmake-gui window.

The rest of the changes are really only relevant to ignition projects that are still in development.

clalancette:
I can also confirm that I can build `gazebo_ros_pkg` on stretch, bionic, and artful with versions 0.4.1, 0.4.0, and 0.6.0 of the `libignition-cmake-dev` packages, respectively.  So I think this should be OK to do.  @j-rivero , @mikaelarguedas , any objections to me merging this?

mikaelarguedas:
:+1: :ship:

j-rivero:
Thanks @clalancette @mxgrey ! LGTM

clalancette:
Merged, and imported.  Thanks!

PR #61, opened 2018-08-10
Import custom gazebo9 for Bionic to solve problems with missing test_fixture lib

j-rivero:
Custom build of gazebo 9.0.0 (version in Bionic repos) with a minimal patch to include static libraries.

Details in the SRU bug:
https://bugs.launchpad.net/ubuntu/+source/gazebo/+bug/1773811

### DIFF
--- a/config/gazebo9_ubuntu_bionic.ppa.yaml
+++ b/config/gazebo9_ubuntu_bionic.ppa.yaml
@@ -1,0 +1,8 @@
+# include the fix for server_fixture while Ubuntu merges the patch:
+# https://bugs.launchpad.net/ubuntu/+source/gazebo/+bug/1773811
+name: gazebo9_import_ubuntu_bionic
+method: http://ppa.launchpad.net/j-rivero/gazebo9bionic/ubuntu
+suites: [bionic]
+component: main
+architectures: [amd64, i386, armhf, arm64, source]
+filter_formula: Package (% *gazebo* )
--- a/config/gazebo9_ubuntu_bionic.ppa.yaml
+++ b/config/gazebo9_ubuntu_bionic.ppa.yaml
@@ -5,4 +5,4 @@ method: http://ppa.launchpad.net/j-rivero/gazebo9bionic/ubuntu
 suites: [bionic]
 component: main
 architectures: [amd64, i386, armhf, arm64, source]
-filter_formula: Package (% *gazebo* )
+filter_formula: Package (% gazebo9-common ) | Package (% gazebo9-doc ) | Package (% gazebo9-plugin-base ) | Package (% gazebo9 ) | Package (% libgazebo9-dev ) | Package (% libgazebo9 )


config/gazebo9_ubuntu_bionic.ppa.yaml
```diff
@@ -0,0 +1,8 @@
+# include the fix for server_fixture while Ubuntu merges the patch:
+# https://bugs.launchpad.net/ubuntu/+source/gazebo/+bug/1773811
+name: gazebo9_import_ubuntu_bionic
+method: http://ppa.launchpad.net/j-rivero/gazebo9bionic/ubuntu
+suites: [bionic]
+component: main
+architectures: [amd64, i386, armhf, arm64, source]
```
mikaelarguedas: As we currently don't ship i386 I would not add it here tpo match the other gazebo9 import rules

config/gazebo9_ubuntu_bionic.ppa.yaml
```diff
@@ -0,0 +1,8 @@
...
+filter_formula: Package (% *gazebo* )
```
mikaelarguedas: The other rules seem to be more specific on the filter regex: https://github.com/ros-infrastructure/reprepro-updater/blob/2f3e9b59aa53601fe32b06b9bb8f6ddbb71c4790/config/gazebo9_ubuntu_artful.upstream.yaml#L6
Can we adapt this one to explicitly list the packages as well ?

config/gazebo9_ubuntu_bionic.ppa.yaml
```diff
@@ -0,0 +1,8 @@
+# include the fix for server_fixture while Ubuntu merges the patch:
+# https://bugs.launchpad.net/ubuntu/+source/gazebo/+bug/1773811
+name: gazebo9_import_ubuntu_bionic
+method: http://ppa.launchpad.net/j-rivero/gazebo9bionic/ubuntu
+suites: [bionic]
+component: main
+architectures: [amd64, i386, armhf, arm64, source]
+filter_formula: Package (% gazebo9-common ) | Package (% gazebo9-doc ) | Package (% gazebo9-plugin-base ) | Package (% gazebo9 ) | Package (% libgazebo9-dev ) | Package (% libgazebo9 )
```
mikaelarguedas: do we need `gazebo9-dbg` as well ? or is it not impacted by the fix ?

j-rivero: the patch ships the static libs, it should not be affected.

nuclearsandwich:
@mikaelarguedas and I talked briefly offline and I wanted to mention since it hasn't come up yet but I think this PR should be targeted against the `refactor` branch in order to be used.

mikaelarguedas:
Yeah I didnt bring it up as it looks like all the recent rules were added only to the master branch (e.g. https://github.com/ros-infrastructure/reprepro-updater/pull/59) so I wanted to check with you and @tfoote how the rules are managed between the 2 branches

nuclearsandwich:
> I wanted to check with you and @tfoote how the rules are managed between the 2 branches

That's a good point. I wonder if the refactor branch is what gets cloned on each import job but the master branch is still in the jenkins home folder. I have an easy way to check.

nuclearsandwich:
> That's a good point. I wonder if the refactor branch is what gets cloned on each import job but the master branch is still in the jenkins home folder. I have an easy way to check.

This doesn't seem to be the case, everything I see is using the refactor branch. @tfoote might know more or it could have been a flub.

j-rivero:
> Is it fair to say that the gazebo 9 packages we imported for stretch and artful at melodic release time don't need the same fix ?

j-rivero:
Arg, sorry just pressed the button to close the PR accidentally.

> Is it fair to say that the gazebo 9 packages we imported for stretch and artful at melodic release time don't need the same fix ?

Since we don't share the packaging information with Debian, yes, it is fair, they were imported from OSRF buildfarm and they should be just fine.

mikaelarguedas:
New packages have been imported on both ROS 1 and ROS 2 farm

PR #64, opened 2019-01-22
add xenial for lark-parser import

dirk-thomas:
As requested on https://github.com/ros2/ros2/issues/638#issuecomment-456473404.


nuclearsandwich:
This package wasn't available on Xenial until [this build](https://build.osrfoundation.org/job/lark-parser-debbuilder/13/) which itself required https://github.com/osrf/lark-parser-release/commit/fa451c48e5feeb54c6dc68d2ca0fc55c2b9ebd41

PR #65, opened 2019-03-21
Skip source architecture before error is raised.

nuclearsandwich:
This doesn't change behavior, source diffs are already omitted because they lack Packages files. This change prints a message that does not contain the word "Exception" before continuing with the sync.

### DIFF
--- a/scripts/sync_ros_packages.py
+++ b/scripts/sync_ros_packages.py
@@ -102,6 +102,9 @@ for ubuntu_distro in distros:
             ubuntu_distro,
             'main',
             package_architecture)
+        if arch == 'source':
+            print("Not computing diff for source architecture.", file=sys.stderr)
+            continue
         try:
             pf_old = diff_repos.get_packagefile_from_url(target_url, name='old')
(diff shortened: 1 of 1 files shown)


scripts/sync_ros_packages.py
```diff
@@ -102,6 +102,9 @@
             ubuntu_distro,
             'main',
             package_architecture)
+        if arch == 'source':
+            print("Not computing diff for source architecture.", file=sys.stderr)
+            continue
```
dirk-thomas: Why not move this block to the beginning of the for loop?

nuclearsandwich: Had to check that nothing in between the diff computation and the start of the block had any desired side-effects. I don't see that they do and so cfe66c8 moves the early continue...earlier.

nuclearsandwich:
So I might have been incorrect about there not being any desired side effects in that loop as Melodic is now missing source packages for its entire testing repository http://repositories.ros.org/status_page/ros_melodic_default.html

PR #66, opened 2019-03-27
Add source to updates_generator before continuing.

nuclearsandwich:
Follow up from #65 which resolves the problem of source packages not being synchronized.

Thanks to @clalancette for reporting the issue.

### DIFF
--- a/scripts/sync_ros_packages.py
+++ b/scripts/sync_ros_packages.py
@@ -78,11 +78,6 @@ dist = conf_params.create_distributions_file(updates_generator)
 
 for ubuntu_distro in distros:
     for arch in arches:
-        # Source architecture doesn't provide a Packages file to diff with.
-        if arch == 'source':
-            print("Not computing diff for source architecture.", file=sys.stderr)
-            continue
-
         d = {'name': 'ros-%s-%s-%s' %
              (options.rosdistro, ubuntu_distro, arch),
              'method': upstream_repo_url,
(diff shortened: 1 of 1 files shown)


nuclearsandwich:
Once this PR merges I think we can restore the testing repository by manually triggering a sync to testing for any affected rosdistros/architectures.

scripts/sync_ros_packages.py
```diff
@@ -94,6 +94,13 @@
 
         updates_generator.add_update_element(conf.UpdateElement(**d))
 
+        # The source architecture doesn't provide a Packages file to diff with.
+        # But it should still be added to the updates_generator above so the
+        # source repository is updated.
+        if arch == 'source':
```
clalancette: This looks OK, but you also need to remove the old check at the top of the loop, I believe.

nuclearsandwich: :man_facepalming:

nuclearsandwich: Fixed via force push of my (apparently) uncommitted changes.

nuclearsandwich:
Repositories that need a sync in order to fix testing:
* [ ] Indigo ARM
* [ ] Indigo Default
* [ ] Kinetic Default
* [ ] Kinetic UXHF
* [ ] Kinetic UXV8
* [ ] Melodic Default
* [ ] Melodic DS
* [ ] Melodic DSV8
* [ ] Melodic UBHF
* [ ] Melodic UBV8

PR #70, opened 2019-09-06
Add import config for ifcfg from packages.osrfoundation.org

claireyywang:
Signed-off-by: claireyywang <clairewang@openrobotics.org>

Move python3-ifcfg from packages.osrfoundation.org to packages.ros2.org.


nuclearsandwich:
@claireyywang thanks for this PR. Can you please add the Debian suites `stretch` and `buster` (which may require building the package for those distros).

In keeping with recent practice I think we'll actually import these packages manually to the ROS bootstrap repos. I think we should still add this import config for visibility but we won't actually use it to deploy the packages.

claireyywang:
Added. Thanks for pointing it out.

PR #51, opened 2018-01-02
Add in bionic to the list of distros.

clalancette:
Signed-off-by: Chris Lalancette <clalancette@openrobotics.org>

PR #72, opened 2020-01-03
Add focal to list of distros

sloretz:
Mimicking #51 

I'm not 100% sure this is still relevant since it sounds like this is a prerequisite to create a `focal` bootstrap repo https://github.com/ros-infrastructure/reprepro-updater/pull/51#issuecomment-354878527, but [one already exists](http://repos.ros.org/repos/ros_bootstrap/dists/focal/).

### DIFF
--- a/src/reprepro_updater/conf.py
+++ b/src/reprepro_updater/conf.py
@@ -35,7 +35,7 @@ import os
 ALL_DISTROS = ['hardy', 'jaunty', 'karmic', 'lucid', 'maverick',
                'natty', 'oneiric', 'precise', 'quantal', 'raring',
                'saucy', 'trusty', 'utopic', 'vivid', 'wily', 'xenial',
-               'yakkety', 'zesty', 'artful', 'bionic',
+               'yakkety', 'zesty', 'artful', 'bionic', 'focal',
                'wheezy', 'jessie', 'stretch', 'buster']
 ALL_ARCHES =  ['amd64', 'i386', 'arm64', 'armel', 'armhf', 'source']
 


sloretz:
> The new bootstrap repository is based on the aptly toolchain and is not using reprepro-updater.
@nuclearsandwich would know better but I think that this branch 'master' is now deprecated fully where the bootstrap repo was the last element using it not the refactor branch.

@tfoote @nuclearsandwich Sounds like I can assume this repo is deprecated and the PR is not needed then. Closing.

nuclearsandwich:
I wanted to check the math on the conclusion here.

This repository contains scripts wrapping reprepro which are still in use on the buildfarm_deployment repository host. The "refactor" branch is in use. ([buildfarm_deployment source](https://github.com/ros-infrastructure/buildfarm_deployment/blob/7e23967b5aa2598cf7347d252773883440e8c20e/modules/profile/manifests/ros/repo.pp#L206-L214)).

The overall reprepro updater configuration which governs what repositories to set up and signing keys and whatnot is taken from the buildfarm_deployment_config ([public example config entry](https://github.com/ros-infrastructure/buildfarm_deployment_config/blob/63c117dc2f9b8bd08bdf0d26a97eff1abebbdeda/hiera/hieradata/buildfarm_role/repo.yaml#L97-L117))

ros_buildfarm has an import_upstream job. The default configuration uses the ros_bootstrap.yml also from the buildfarm_deployment_config. ([public example config entry](https://github.com/ros-infrastructure/buildfarm_deployment_config/blob/63c117dc2f9b8bd08bdf0d26a97eff1abebbdeda/hiera/hieradata/buildfarm_role/repo.yaml#L119-L131))

The config files in this repository for importing from various upstreams are technically usable from the import_upstream job but doing so is quite opaque, especially to people setting up buildfarm replicas. I think at this point we've settled on importing everything into the aptly-backed bootstrap repository and using the import_upstream job to get whatever's in that to the ROS and ROS 2 repositories.

tfoote:
My conclusion was that `master` was deprecated. We're still using `refactor` branch. Though we could consider moving the wrapper scripts into the buildfarm now as that's the only think we're using these days and we've already got lots of wrapper scripts in ros_buildfarm there's not a real need for them to be in a separate repository. 

As we're switching to the new repository management tool we should find a way to document and track those packages natively in their old home instead of trying to maintain these rules that we can't/won't ever execute.

PR #75, opened 2020-04-29
Add gazebo11 Debian Buster and Ubuntu Focal

j-rivero:
For both, Noetic (ROS1) and Foxy (ROS2) we need to import gazebo11 packages together with all its dependencies. I added two files here to import all the needed files from packages.osrfoundation.org. 

Some packages are affected by https://github.com/ignition-tooling/release-tools/issues/186 and don't have -dbg packages. I think this should not stop us from starting the upload in order to unblock the `gazebo_ros_pkgs` porting, testing and release.

/cc @chapulina

PR #76, opened 2020-05-18
Updated configuration for ignition-citadel and gazebo11.

nuclearsandwich:
Followup to #75 to reflect what was actually imported.


config/ignition_citadel_gazebo11_ubuntu_focal.yaml
```diff
@@ -0,0 +1,6 @@
...
...3-gpu-lidar-dev) | Package (% libignition-sensors3-imu) | Package (% libignition-sensors3-imu-dev) | Package (% libignition-sensors3-lidar) | Package (% libignition-sensors3-lidar-dev) | Package (% libignition-sensors3-logical-camera) | Package (% libignition-sensors3-logical-camera-dev) | Package (% libignition-sensors3-magnetometer) | Package (% libignition-sensors3-magnetometer-dev) | Package (% libignition-sensors3-rendering) | Package (% libignition-sensors3-rendering-dev) | Package (% libignition-sensors3-rgbd-camera) | Package (% libignition-sensors3-rgbd-camera-dev) | Package (% libignition-sensors3-thermal-camera) | Package (% libignition-sensors3-thermal-camera-dev) | Package (% libignition-tools-dev) | Package (% libignition-transport8) | Package (% libignition-transport8-core-dev) | Package (% libignition-transport8-dbg) | Package (% libignition-transport8-dev) | Package (% libignition-transport8-log) | Package (% libignition-transport8-log-dev) | Package (% libogre-2.1) | Package (% libogre-2.1-dev) | Package (% libsdformat9) | Package (% libsdformat9-dbg) | Package (% libsdformat9-dev) | Package (% sdformat9-dbg) | Package (% sdformat9-doc) | Package (% sdformat9-sdf)
```
j-rivero, line 6: cmake2 is available from Ubuntu repositories. We don't need to import it.

nuclearsandwich: Because it's not present in the upstream (packages.osrfoundation.org) repository and because the filter won't fail if one of the items is not found, I'm super torn between keeping the filter_formula the same between the Debian and Ubuntu to make upkeep of this list easier and not having a package which you rightfully identify as spurious in the list here.

nuclearsandwich: Removed in 68bfa05

config/ignition_citadel_gazebo11_ubuntu_focal.yaml
```diff
@@ -0,0 +1,6 @@
...
```
chapulina, line 6: This is missing `libignition-cmake2-dev`, but buster has it.

PR #77, opened 2020-09-04
Add Ignition Physics TPE packages

chapulina:
Follow-up to #76. Added 4 TPE packages that I didn't realize hadn't been included before:

`Package (% libignition-physics2-tpe) | Package (% libignition-physics2-tpe-dev) | Package (% libignition-physics2-tpelib) | Package (% libignition-physics2-tpelib-dev)`

You can see the original ones here:

* http://packages.osrfoundation.org/gazebo/debian-stable/pool/main/i/ignition-physics2/
* http://packages.osrfoundation.org/gazebo/ubuntu-stable/pool/main/i/ignition-physics2/

PR #78, opened 2020-09-29
add colcon upstream config

dirk-thomas:
A starting point to configure the import from the `colcon` upstream repo https://packagecloud.io/dirk-thomas/colcon.

Maybe we want to state version numbers of packages individually.


config/colcon.upstream.yaml
```diff
@@ -0,0 +1,6 @@
+name: colcon
+method: https://packagecloud.io/dirk-thomas/colcon
```
nuclearsandwich, line 2:
```suggestion
method: https://packagecloud.io/dirk-thomas/colcon/ubuntu
```

The actual repository is distribution specifc so the above will work for Ubuntu distributions but not Debian distributions suchas Buster and Stretch.

config/colcon.upstream.yaml
```diff
@@ -0,0 +1,6 @@
...
+suites: [xenial, bionic, focal, stretch, buster]
+component: main
+architectures: [i386, amd64, armhf, arm64, source]
+filter_formula: Package (% python3-colcon-* )
```
nuclearsandwich, line 6:
```suggestion
filter_formula: (Package (% python3-colcon-argcomplete, $Version (% 0.3.3*))|(Package (% python3-colcon-bash, $Version (% 0.4.2*))|(Package (% python3-colcon-bazel, $Version (% 0.1.0*))|(Package (% python3-colcon-cd, $Version (% 0.1.1*))|(Package (% python3-colcon-cmake, $Version (% 0.2.25*))|(Package (% python3-colcon-cmake, $Version (% 0.2.26*))|(Package (% python3-colcon-common-extensions, $Version (% 0.2.1*))|(Package (% python3-colcon-core, $Version (% 0.6.0*))|(Package (% python3-colcon-coveragepy-result, $Version (% 0.0.8*))|(Package (% python3-colcon-defaults, $Version (% 0.2.5*))|(Package (% python3-colcon-devtools, $Version (% 0.2.2*))|(Package (% python3-colcon-ed, $Version (% 0.1.0*))|(Package (% python3-colcon-lcov-result, $Version (% 0.4.1*))|(Package (% python3-colcon-library-path, $Version (% 0.2.1*))|(Package (% python3-colcon-metadata, $Version (% 0.2.5*))|(Package (% python3-colcon-mixin, $Version (% 0.2.0*))|(Package (% python3-colcon-notification, $Version (% 0.2.13*))|(Package (% python3-colcon-output, $Version (% 0.2.11*))|(Package (% python3-colcon-package-information, $Version (% 0.3.3*))|(Package (% python3-colcon-package-selection, $Version (% 0.2.8*))|(Package (% python3-colcon-package-selection, $Version (% 0.2.9*))|(Package (% python3-colcon-parallel-executor, $Version (% 0.2.4*))|(Package (% python3-colcon-pkg-config, $Version (% 0.1.0*))|(Package (% python3-colcon-powershell, $Version (% 0.3.6*))|(Package (% python3-colcon-python-setup-py, $Version (% 0.2.6*))|(Package (% python3-colcon-recursive-crawl, $Version (% 0.2.1*))|(Package (% python3-colcon-ros, $Version (% 0.3.19*))|(Package (% python3-colcon-ros, $Version (% 0.3.20*))|(Package (% python3-colcon-test-result, $Version (% 0.3.8*))|(Package (% python3-colcon-zsh, $Version (% 0.4.0*))
```

This expanded query includes each package in the colcon repository currently along with it's version (with a wildcard for the package revision). With this change, future updates to packages in the colcon repository would open a PR adding additional packages or changing the version of existing packages to signal a desire to update the ROS bootstrap repository.

nuclearsandwich (commented):
@dirk-thomas as we discussed offline, using specific package names and version numbers in the upstream filter makes reproducing the bootstrap repository more consistent. Colcon maintainers can open patches adding additional packages or updating package version (I left the package revision wild-carded but we could specify that as well).

There is one other issue I noticed when testing this, the Debian and Ubuntu repositories use different URLs.

We could, like the Ignition #77 PRs split into two separate configs for Debian and Ubuntu or if you wanted to cheat for the time being in order to retain single file edits I can make the adjustment directly in our mirror configuration so that *it* does the appropriate thing. However that will mean that anyone using reprepro-updater and this config will encounter errors if they try to use this config completely.

config/colcon.upstream.yaml
dirk-thomas: Is there a way to add Debian without duplicating the whole file?

config/colcon.upstream.yaml
dirk-thomas: Is there a way to add newlines into these formula?

config/colcon.upstream.yaml
nuclearsandwich: Not that I know of in this configuration structure.

config/colcon.upstream.yaml
nuclearsandwich: > Is there a way to add newlines into these formula?

Not in a way that guarantees it will still work as a reprepro-updater configuration since this filter is passed as a command line argument.

dirk-thomas: I hope this works: bd8982c47d35d71f3cead2d5fd708b07721de86f.

config/colcon.upstream.yaml
dirk-thomas: Split and updated the URLs / suites accordingly: 2f37b6937ff499c0fcd05360f014371f8db7d224.

config/colcon.upstream.yaml
nuclearsandwich: That seems like it should.

nuclearsandwich (approved):

nuclearsandwich:
I've started the work of getting this imported on the bootstrap repository. I will link import_upstream urls in this PR discussion when they're up.

nuclearsandwich:
> I've started the work of getting this imported on the bootstrap repository. I will link import_upstream urls in this PR discussion when they're up.

I should have clarified: I've got the prep-work done so these changes are staged but I wasn't going to perform the import until we merge this PR.

dirk-thomas:
I removed duplicates in 1dbd8285069b8c330380a4a090b258380cd038a3 to only list the latest version of each package.

dirk-thomas:
I also made sure that `python3-colcon-bazel` is also available for Focal.

nuclearsandwich (approved):

nuclearsandwich:
Import jobs

http://build.ros.org/job/import_upstream/134/
http://build.ros2.org/job/import_upstream/290/

PR #81, opened 2020-10-09
Update catkin_lint upstream to new PPA

nuclearsandwich:
Per discussion in https://github.com/ros/rosdistro/issues/24251

### DIFF
--- a/config/catkin_lint.upstream.yaml
+++ b/config/catkin_lint.upstream.yaml
@@ -3,4 +3,4 @@ method: http://ppa.launchpad.net/roehling/ros/ubuntu
 suites: [xenial, bionic, focal]
 component: main
 architectures: [i386, amd64, armhf, arm64, source]
-filter_formula: Source (== catkin-lint), $Version (% 1.6.10*)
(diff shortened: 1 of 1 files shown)


roehling (commented):
Trusty is EOL since April 2019 and no longer part of the PPA.

If you want to lock to a specific version, I suggest 1.6.10-* as wildcard (include the dash), otherwise .1* might also match .10 to .19

nuclearsandwich:
I've removed trusty from the import configuration.

roehling (approved):
LGTM

nuclearsandwich:
> If you want to lock to a specific version, I suggest 1.6.10-* as wildcard (include the dash), otherwise .1* might also match .10 to .19

As written it was `.10*` which using the glob format would match `100` - `109`. as well as .10. So adding the dash is still an improvement.

gavanderhoorn:
Thanks @nuclearsandwich :+1: 

And @roehling :+1:

nuclearsandwich:
Import jobs: 
* http://build.ros.org/job/import_upstream/137/
* http://build.ros2.org/job/import_upstream/293/

PR #83, opened 2020-10-26
add vcstool upstream config

dirk-thomas:
To configure the import from the `vcstool` upstream repo https://packagecloud.io/dirk-thomas/vcstool.

This config only contains the Python 2 packages. I intend to do one last Python 2 release directly into the ROS apt repos adding a message that this is the last Python 2 release and users should update to Python 3 packages.

### DIFF
--- a/config/vcstool.debian.upstream.yaml
+++ b/config/vcstool.debian.upstream.yaml
@@ -1,0 +1,6 @@
+name: vcstool
+method: https://packagecloud.io/dirk-thomas/vcstool/debian
+suites: [jessie, stretch, buster]
+component: main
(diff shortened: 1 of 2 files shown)


dirk-thomas:
With dirk-thomas/vcstool#176 future releases will be made to packagecloud.io (and PyPI as before).

dirk-thomas:
@nuclearsandwich Friendly ping.

nuclearsandwich (approved):

nuclearsandwich:
Import jobs
* http://build.ros.org/job/import_upstream/140/
* http://build.ros2.org/job/import_upstream/296/

PR #85, opened 2020-10-29
Use NamedTemporaryFile to avoid problems with PackageFile+StringIO in python3

j-rivero:
`PackageFile` constructor:
```python
def __init__(self, name, file_obj=None, encoding="utf-8"):
        """Creates a new package file object.
         name - the name of the file the data comes from
         file_obj - an alternate data source; the default is to open the
                       file with the indicated name.
          """
           if file_obj is None:
               file_obj = open(name, 'rb')
```
and iter as:
```python
        def __iter__(self):
           line = self.file.readline().decode(self.encoding)
```
When using `StringIO` the readline returns `str` and the decode fails. When using real filesystem paths, the readline returns `bytes` and the thing seems to work fine.

Upstream code could be patched to work with `StringIO` I'll see if I have time to send a patch there.

Tested here:
[Build Status](https://build.test.ros2.org/job/Erel_sync-packages-to-testing_bionic_amd64/14/)

### DIFF
--- a/scripts/diff_packages.py
+++ b/scripts/diff_packages.py
@@ -33,12 +33,12 @@ def main():
     )
 
     try:
-        pf_old = diff_repos.get_packagefile_from_url(target_url, name='old')
+        pf_old = diff_repos.get_packagefile_from_url(target_url)
     except RuntimeError as ex:
         parser.error("%s" % ex)
 
     try:
-        pf_new = diff_repos.get_packagefile_from_url(upstream_url, name='new')
+        pf_new = diff_repos.get_packagefile_from_url(upstream_url)
     except RuntimeError as ex:
         parser.error("%s" % ex)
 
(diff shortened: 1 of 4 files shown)


j-rivero:
Upstreamed: https://salsa.debian.org/python-debian-team/python-debian/-/merge_requests/30

nuclearsandwich:
[Build Status](http://build.ros2.org/job/Drel_sync-packages-to-testing_bionic_amd64/433/) is a test build of this on build.ros2.org to make sure it doesn't cause any regressions when the scripts are run with Python 2.

nuclearsandwich (approved):
I was, at first, concerned about losing the explicit names of old and new but there isn't anywhere they show up in user-facing content.

tfoote (approved):

ros-discourse:
This pull request has been mentioned on **ROS Discourse**. There might be relevant details there:

https://discourse.ros.org/t/changes-to-ros-buildfarm-affecting-repository-import-jobs/17047/1

PR #91, opened 2021-01-05
Increase versions of OSRF packages

chapulina:
Updated style to make the file easier to read and pinned to specific versions - they match what's latest in packages.osrfoundation.org.

PR #93, opened 2021-01-15
Update Ignition versions and fix libogre-dev

chapulina:
Follow-up to https://github.com/ros-infrastructure/reprepro-updater/pull/91

Fix `libogre2.1-dev`.

There are some new releases of other libraries too, and the older version is not around, so we need to point to the newer versions.


nuclearsandwich (approved):

nuclearsandwich:
Import jobs for ROS https://build.ros.org/job/import_upstream/149/ and ROS 2 https://build.ros2.org/job/import_upstream/305/

PR #94, opened 2021-02-06
Add the pyside2 source package to the import formula.

nuclearsandwich:
Because the source package for the pyside2 collection is called simply `pyside2` it was not imported as was originally intended along with the rest of these packages. Because the source package for shiboken2 is called `shiboken2` that source package was successfully imported previously.

The version of pyside2 which was imported previously is no longer available in the linked PPA so this import should not be re-run unless the PPA is updated or an examination of the differences between the current ROS and PPA packages is performed.

PR #95, opened 2021-02-18
Add gazebo9 source package import formula (Bionic)

j-rivero:
The formula will import the gazebo source package corresponding to the packages in http://repos.ros.org/repos/ros_bootstrap/pool/main/g/gazebo/.

There is an ancient 1.9 source package named gazebo there that can be overwritten safely since all versions using it are EOL.

### DIFF
--- a/config/gazebo9_bionic.yaml
+++ b/config/gazebo9_bionic.yaml
@@ -1,5 +1,5 @@
 name: gazebo9bionic
-method: https://launchpad.net/~j-rivero/+archive/ubuntu/gazebo9bionic
+method: http://ppa.launchpad.net/j-rivero/gazebo9bionic/ubuntu
 suites: [bionic]
 component: main
 architectures: [source]
--- a/config/gazebo9_bionic.yaml
+++ b/config/gazebo9_bionic.yaml
@@ -4,7 +4,8 @@ suites: [bionic]
 component: main
 architectures: [amd64, arm64, armhf, i386, source]
 filter_formula: "\ 
-((Package (= gazebo9) |\
+((Package (= gazebo) |\
+(Package (= gazebo9) |\
 Package (= gazebo9-common) |\
 Package (= gazebo9-doc) |\
 Package (= gazebo9-plugin-base) |\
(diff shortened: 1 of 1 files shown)


config/gazebo9_bionic.yaml
```diff
@@ -0,0 +1,6 @@
+name: gazebo9bionic
+method: https://launchpad.net/~j-rivero/+archive/ubuntu/gazebo9bionic
```
nuclearsandwich, line 2:
```suggestion
method: http://ppa.launchpad.net/j-rivero/gazebo9bionic/ubuntu
```

The method needs to be the apt repository URL not the web link to the PPA.

config/gazebo9_bionic.yaml
```diff
@@ -0,0 +1,6 @@
...
+suites: [bionic]
+component: main
+architectures: [source]
```
nuclearsandwich, line 5: Do you think it makes sense to import the binary packages as well in order to ensure that we're soundly using packages generated from this source package? That's what we did for pyside2 and I think it makes sense to do here as well.

config/gazebo9_bionic.yaml
j-rivero: ouch, thanks.

nuclearsandwich (approved):

config/gazebo9_bionic.yaml
```diff
@@ -0,0 +1,6 @@
+name: gazebo9bionic
+method: http://ppa.launchpad.net/j-rivero/gazebo9bionic/ubuntu
+suites: [bionic]
+component: main
+architectures: [amd64, arm64, armhf, i386, source]
+filter_formula: Package (% gazebo9), $Version (% 9.0.0+dfsg5-3ubuntu1+ppa2)
```
nuclearsandwich, line 6: Do we need the libgazebo9* packages as well? Would you accept a change that uses exact package names for clarity, matching the format used in #94?

j-rivero: 2e1b8d0

nuclearsandwich (approved):

nuclearsandwich:
Import jobs
https://build.ros.org/job/import_upstream/156/
https://build.ros2.org/job/import_upstream/309/

PR #96, opened 2021-02-19
Add opensplice69 source package import formula (Bionic)

j-rivero:
The formula will import the opensplice69 packages (including source package) corresponding to the rebuild effort of the same source used in http://repos.ros.org/repos/ros_bootstrap/pool/main/libo/libopensplice69/ to recreate the source package file.

Revision number has been bumped to `6.9.190705+osrf1-2~bionic~bionic` (duplicate ~bionic is an error but I think it does not affect in anyway the expected work of the packages).

I ran the debdiff tool on both packages and try to run auto-abi checker: https://github.com/j-rivero/bootstrap_abi_checking/runs/1929361098?check_suite_focus=true

The auto-abi checker failed (need customization) but the debdiff tool is reporting:
```
2021-02-18T18:08:01.5303048Z #14 1.337 File lists identical (after any substitutions)
2021-02-18T18:08:01.5303606Z #14 1.358 
2021-02-18T18:08:01.5304205Z #14 1.358 Control files: lines which differ (wdiff format)
2021-02-18T18:08:01.5335032Z #14 1.358 ------------------------------------------------
2021-02-18T18:08:01.5335837Z #14 1.358 Installed-Size: [-22581-] {+22673+}
2021-02-18T18:08:01.5336685Z #14 1.358 Version: [-6.9.190705+osrf1-1~bionic-] {+6.9.190705+osrf1-2~bionic~bionic+}
2021-02-18T18:08:01.6803576Z #14 DONE 1.4s
``` 
I think we should be safe to import.

PR #97, opened 2021-02-22
Add opensplice69 source package import formula (Xenial)

j-rivero:
Twin of pull #96 for Xenial


nuclearsandwich:
I'm going to combine this and #96 into a single config file and merge them together.

PR #99, opened 2021-02-22
Add the pyside2 source package to the import formula (Bionic)

j-rivero:
Packages in the ROS repository has no source package but it is available in the PPA. Import it from there to complete the collection.


nuclearsandwich:
Since this package is not used by any ROS distribution, active or otherwise we removed it from the bootstrap repository. https://discourse.ros.org/t/removing-packages-from-ros-bootstrap-repository/19386

PR #102, opened 2021-03-01
Add ignition source packages import formula. Latest versions for launch2, gazebo3, rendering3, sensors3, sdformat9 (Focal)

j-rivero:
Ignition Focal packages in the bootstrap repository lack of source packages for Focal.

~~Packages are split in two sets:~~
 * ~~Some of the versions did not change in the stable `packages.osrfoundation.org` repository, import source packages for them in `ignition_focal_source.yaml`~~
 * ~~Other versions were updated and the source package was lost after the update. I've rebuilt the same code and upload these old versions to the testing server in `ignition_focal_from_testing.yaml`~~

**Update:** one single file containing:
 * Missing source packages for the latest versions
 * Update  to latest version together with source packages for old versions.


j-rivero:
@chapulina quick note that with this PR we start to change the behavior and latest Ignitions land in the ROS repository. I assume that we want this sync, will probably get many new users of these versions.

chapulina:
> we start to change the behavior and latest Ignitions land in the ROS repository

I'm not sure I understand, will the process be automated now? What's the change in behaviour?

j-rivero:
> I'm not sure I understand, will the process be automated now? What's the change in behaviour?

Sorry, the change would be trying to be more in sync between the ROS repo and the Ignition repo. It's not going to be fully automated but tools to make things easier are being merged.

chapulina:
Ah gotcha. Upgrading the Citadel versions of all libraries on packages.ros.org should be safe and good to do at any point. We just need to be sure to update them as a whole.

nuclearsandwich (approved):
@j-rivero I'd be glad if you gave this the once-over after all the commits I added to it.

config/ignition_citadel_gazebo11_ubuntu_focal.yaml
```diff
@@ -1,106 +1,93 @@
-name: gazebo11_ubuntu_focal.yaml
+name: ignition_focal_source
```
j-rivero, line 1: Not sure if changing the name impacts on your workflow @nuclearsandwich . I'm trying to be accurate, if possible, if not, feel free to drop the suggestion.

```suggestion
name: ignition_citadel_gazebo11_focal_source
```

j-rivero:
> @j-rivero I'd be glad if you gave this the once-over after all the commits I added to it.

Thanks for fighting making that format somehow readable, looks like an improvement to me. The meaningful commits helped to review. All ready, feel free to merge. I just left a minor suggestion about the name.

config/ignition_citadel_gazebo11_ubuntu_focal.yaml
nuclearsandwich: While there's no hard requirement. The name should definitely match the filename. Good catch.

PR #104, opened 2021-03-04
Converge lark-parser packages on a single orig.tar for 0.7.2

nuclearsandwich:
When importing the new source package for Xenial it introduced a conflict with the already-present source package orig.tar files so we've updated all of them to use the same one.

### DIFF
--- a/config/lark-parser_stretch_buster.yaml
+++ b/config/lark-parser_stretch_buster.yaml
@@ -1,0 +1,12 @@
+name: lark-parser_stretch_buster
+method: http://packages.osrfoundation.org/gazebo/debian-testing
+suites: [buster, stretch]
+component: main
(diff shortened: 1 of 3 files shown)


j-rivero:
I've split the file into Ubuntu/Debian. The packages for Debian are in place in the testing repository. All should be ready.

j-rivero (approved):

nuclearsandwich:
> I've split the file into Ubuntu/Debian. The packages for Debian are in place in the testing repository. All should be ready.

Thank you @j-rivero!

nuclearsandwich:
Import upstream jobs:
* build.ros.org https://build.ros.org/job/import_upstream/157/
* build.ros2.org https://build.ros2.org/job/import_upstream/310/

PR #105, opened 2021-03-19
aptly automation for reprepro updater

j-rivero:
The PR implements `aptly` automation to be able to process and publish a configuration file into the existing repositories and publications of `aptly` in the server. Some assumptions: 

 * Script is currently programmed to work with one of the configuration files in the `config` directory. Most of the PRs of this repository are about adding or modifying configuration files.
 * The configurations files uses filtering from reprepro syntax. I made little conversion to make it work, the [conversion might need some more tweaks](https://github.com/ros-infrastructure/reprepro-updater/compare/master...j-rivero:aptly_scripts?expand=1#diff-426629ab4bb8f5cc170246f0e0ccb1a23890962912aa239e05eb8b860136d660R12-R20).
 * The [main logic of the whole process is easily readable](https://github.com/ros-infrastructure/reprepro-updater/compare/master...j-rivero:aptly_scripts?expand=1#diff-426629ab4bb8f5cc170246f0e0ccb1a23890962912aa239e05eb8b860136d660R226-R243) from the `run`  method of the UpdaterManager class. Summarized:
  1. check that mirrors machting the name name of the config file does not exist
  1. for every distribution in the config file:
     1. create an aptly mirror with the info from config file
     2. check that there is a source package imported in the mirror
     3. import mirror into the existing local repository
     4. create an snapshot with timestamp from local repository
     5. publish the snapshot into the existing publications

 * The tool assumes the different naming conventions in the server:
   * [Aptly mirrors from config file use the form: `${name_in_config_file}-${distro}`](https://github.com/ros-infrastructure/reprepro-updater/compare/master...j-rivero:aptly_scripts?expand=1#diff-426629ab4bb8f5cc170246f0e0ccb1a23890962912aa239e05eb8b860136d660R184-R185)
   * [Aptly repositories in the server are named `ros_bootstrap-${distro}`](https://github.com/ros-infrastructure/reprepro-updater/compare/master...j-rivero:aptly_scripts?expand=1#diff-426629ab4bb8f5cc170246f0e0ccb1a23890962912aa239e05eb8b860136d660R187-R188),
   * [Aptly publication endpoins are all distributions under `filesystem:live:ros_bootstrap`](https://github.com/ros-infrastructure/reprepro-updater/compare/master...j-rivero:aptly_scripts?expand=1#diff-426629ab4bb8f5cc170246f0e0ccb1a23890962912aa239e05eb8b860136d660R181-R182), which is defined in `aptly.conf` (tests are recreating this configuration).

 * If I'm not wrong there is a change from current bash scripts: the scripts go over all repositories, create an snapshot and publish them while the approach of this PR is only doing it for the repositories affected by the configuration file processed.
 
 * At least one source package needs to be present in the configuration file in order to proceed with the publications. [The check](https://github.com/ros-infrastructure/reprepro-updater/compare/master...j-rivero:aptly_scripts?expand=1#diff-426629ab4bb8f5cc170246f0e0ccb1a23890962912aa239e05eb8b860136d660R62-R72) is based on finding a `_source`  postfix in one of the packages available from the config file.

The tool currently does not provide any log or message of feedback to the user if there are no errors. Probably is a good idea to add some messages about what's going on. I'm not sure about the location of the files in this repo or if we want to remove the other scripts or if they are still in use.

The PR includes a good number of tests and example files and filters together with github action integration.

Future work: make the tool to run on new or modified files added by PRs to check that they can work nicely in format and content with the tool.

**Update:**

 * The tool now displays information about any step being performed.
 * Thanks to @nuclearsandwich the tool check all the new packages in mirrors looking for all source packages declared by all the binary packages instead of assuming that finding one is enough.


nuclearsandwich (commented):
I had a comment trapped in "pending" because I stopped using the GitHub interface and opened my own subordinate PR.

scripts/aptly/aptly_importer.py
```diff
@@ -0,0 +1,288 @@
...
+from sys import stderr, exit
+import re
+import time
+import yaml
+
+
+class Reprepro2AptlyFilter():
+    def convert(self, filter_str):
+        # remove begin/end whitespaces
+        r_str = filter_str.lstrip()
```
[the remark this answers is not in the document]
j-rivero: oh yes, the intention was to clean up all whitespaces left and right.

j-rivero:
Sorry, just found another problem handling errors, fixed in  1d6a1f9

scripts/aptly/aptly_importer.py
```diff
@@ -0,0 +1,338 @@
...
+
+    def __create_aptly_snapshot(self, distribution):
+        self.__log('Creating an aptly snapshot from local aptly repository')
+        self.aptly.run(['snapshot', 'create', self.__get_snapshot_name(distribution),
+                        'from', 'mirror', self.__get_mirror_name(distribution)])
+        self.__log_ok(f"snapshot {self.__get_snapshot_name(distribution)} created from mirror {self.__get_mirror_name(distribution)}")
+
+    def __error(self, msg):
+        print(f"Update Manager error: {msg} \n", file=stderr)
+        exit(-1)
```
nuclearsandwich, line 235: Does it make sense to use a low valued exit code rather than -1/255?

scripts/aptly/aptly_importer.py
```diff
@@ -0,0 +1,338 @@
...
+                self.config['filter_formula'])
+            self.method = self.config['method']
+            self.name = self.config['name']
+            self.suites = self.config['suites']
+        except KeyError as e:
+            self.__error(f"{e} key was not found in file {input_file}")
+
+    def __error(self, msg):
+        print(f"Configuration file error: {msg} \n", file=stderr)
+        exit(-1)
```
nuclearsandwich, line 183: Does it make sense to use a low valued exit code rather than -1/255?

scripts/aptly/aptly_importer.py
```diff
@@ -0,0 +1,338 @@
...
+
+    def __init__(self, debug=False, config_file=None):
+        self.debug = debug
+        self.config_file = config_file
+
+    def __error(self, cmd, msg, exit_on_errors=False):
+        print(f"Aptly error running: {cmd}", file=stderr)
+        print(f"  --> {msg} \n", file=stderr)
+        if exit_on_errors:
+            exit(-1)
```
nuclearsandwich, line 39: Does it make sense to use a low valued exit code rather than -1/255?

nuclearsandwich (commented):
Overall this is looking solid.

One interesting quirk of this repository is that the scripts used by the ros_buildfarm are on the `refactor` branch rather than master.

The config folders have largely stayed on master because that's where the PRs are opened by default. We might have an offline discussion about where to land these scripts in the near future but they don't break anything by landing on master first.

scripts/aptly/aptly_importer.py
j-rivero: Used 1 for generic errors, 2 for config file errors. 7e74b46

scripts/aptly/aptly_importer.py
j-rivero: Used 1 for generic errors, 2 for config file errors. 7e74b46

scripts/aptly/aptly_importer.py
j-rivero: Used 1 for generic errors, 2 for config file errors. 7e74b46

j-rivero:
> One interesting quirk of this repository is that the scripts used by the ros_buildfarm are on the `refactor` branch rather than master.
> 
> The config folders have largely stayed on master because that's where the PRs are opened by default. We might have an offline discussion about where to land these scripts in the near future but they don't break anything by landing on master first.

+1 to clean up the whole thing

j-rivero:
One final thing: the test is potentially dangerous in the case of being run in the real server. It can destroy permanent repos or endpoints. Not sure what could be the best way of preventing the disaster. @nuclearsandwich any idea?

j-rivero:
> One final thing: the test is potentially dangerous in the case of being run in the real server. It can destroy permanent repos or endpoints. Not sure what could be the best way of preventing the disaster. @nuclearsandwich any idea?

I added an stupid safe check in 47f400a should be enough not to kill the server if the test is run.

nuclearsandwich (changes requested):
I tried this out pre-testing #102.

The first issue I saw is that it can't handle mirrors already existing.
One quick workaround would be to delete and recreate mirrors on each run but if the mirror has a lot of packages to index that could slow down the update.

The second issue is that I think by default we should omit the snapshot and publish steps so that we can use the existing `snapshot-and-publish-all` script, potentially rewriting that in Python to handle errors better.

The reason for creating snapshots for all distros instead of just the changed ones is that snapshots don't cost much to store and it allows us to rollback monolithically instead of having to identify what snapshots need to be rolled back for each distribution.

I'll push some changes making the snapshot and publish step conditional and may try to take a crack at the mirror issue as well.

Thanks for this work. Reading and editing the script has been really fluid and I love the testing options.

scripts/aptly/aptly_importer.py
```diff
@@ -0,0 +1,338 @@
...
+        return True
+
+
+def main():
+    """
+    Usage: python3 aptly_importer.py <config_file>
+    """
+    usage = "usage: %prog config_file"
+    parser = argparse.ArgumentParser(usage)
+    parser.add_argument('config_file', type=str, nargs='+', default=None)
```
nuclearsandwich, line 326: Here nargs are set but only the first config_file is passed to the UpdaterManager. Is there a plan for handling more config files with multiple update managers or should we remove the nargs option (I'm fine with that to keep thinks simpler).

j-rivero:
> The first issue I saw is that it can't handle mirrors already existing.
> One quick workaround would be to delete and recreate mirrors on each run but if the mirror has a lot of packages to index that could slow down the update.

I think that changes here are easy to made. We just need to decide which is the best way of handling existing repositories, if delete and recreate, respect them and run against the current state (that might be unstable since it introduce unexpected statuses).

> The second issue is that I think by default we should omit the snapshot and publish steps so that we can use the existing `snapshot-and-publish-all` script, potentially rewriting that in Python to handle errors better.

Please check https://github.com/j-rivero/reprepro-updater/pull/2 . There is an option of only using mirrors there.

> The reason for creating snapshots for all distros instead of just the changed ones is that snapshots don't cost much to store and it allows us to rollback monolithically instead of having to identify what snapshots need to be rolled back for each distribution.

Aaaah ok. I made the decision of only going snapshot and publish the modified repositories.  We can change it.

nuclearsandwich:
>  Not sure what could be the best way of preventing the disaster. @nuclearsandwich any idea?

For my own understanding you mean that running aptly_importer_TEST.py could result in destructive action on a production server?

If that's the case either the whole file or the destructive tests should probably assert that some dramatically named environment variable like `ALLOW_DESTRUCTIVE_TESTS` is defined before continuing.

j-rivero:
> If that's the case either the whole file or the destructive tests should probably assert that some dramatically named environment variable like `ALLOW_DESTRUCTIVE_TESTS` is defined before continuing.

b1c1224 and fixed logic here 98ea172

scripts/aptly/aptly_importer.py
j-rivero: good catch, that was a mistake. Removed in 0069bd7

nuclearsandwich (approved):
Once we get the tests behaving this is great!

PR #107, opened 2021-03-31
Use aptly importer to test changes in config files in PRs

j-rivero:
The PR implements a mechanism to add CI testing to PRs for configuration files.

Two new arguments has been added to the script:
 * **--ignore-signatures** this helps the CI to import from anywhere. It could use to force imports into the bootstrap if needed at some point.
 * **--only-mirror-creation**: this makes the script to only run the steps that handle the mirror creation. No repository, endpoints or snapshots are touched.
                
Github actions are refactor to share comm installation steps. Adding a `CONFIG_TESTER` job that check for changing files in the commit under testing and run  `aptly_importer.py` in all them with the option of only creating the mirrors. The CI can be seen in action in the [last commit of this PR](https://github.com/j-rivero/reprepro-updater/runs/2204558834).

There are other changes like improving the how test file directory is being set or changes in config files used during testing.


scripts/aptly/aptly_importer.py
```diff
@@ -321,11 +338,19 @@ def run(self):
 
 def main():
     """
-    Usage: python3 aptly_importer.py [--simulate-repo-import] <config_file>
+    Usage: python3 aptly_importer.py [parameters] <config_file>
```
nuclearsandwich, line 341: Does it make sense to preserve the explicit parameters in the usage text for quick reference?

PR #109, opened 2021-04-15
Ignition Edifice

chapulina:
Part of https://github.com/ros/rosdistro/issues/28960

What I did:

* Copied files from Citadel
* Removed Gazebo 11, as well as libraries that are common between Citadel and Edifice (`cmake2`, `math6`, `tools`, `plugin1`)
* Bumped the versions for all the others

Some things I noticed, but wasn't sure how to handle:

* Focal is not including the `ignition-citadel` metapackage, so I didn't include `ignition-edifice`. I see the package has been uploaded though: http://packages.ros.org/ros/ubuntu/pool/main/i/ignition-citadel/. I wonder if the config file is wrong and the metapackage has been uploaded in an earlier version. Should I include `ignition-edifice` this time?
* The format `Package, Version | Package, Version` is used in some places, and the format `Package  Package, Version` is used in others. Which one is preferred? The 2nd seems easier to maintain.
* There's a mismatch of sub-packages between Focal and Buster sometimes. For example, there are 6 packages for `transport10` on Buster and only one on Focal. Should all 6 be added to Focal, or is there some globbing going on?


chapulina:
I'm not sure why `Aptly Importer / Bionic Aptly CONFIG_TESTER - debug true` failed

nuclearsandwich:
>     * Focal is not including the `ignition-citadel` metapackage, so I didn't include `ignition-edifice`. I see the package has been uploaded though: [http://packages.ros.org/ros/ubuntu/pool/main/i/ignition-citadel](http://packages.ros.org/ros/ubuntu/pool/main/i/ignition-citadel/). I wonder if the config file is wrong and the metapackage has been uploaded in an earlier version. Should I include `ignition-edifice` this time?

Especially if there is a rosdep entry for the metapackage, let's make sure it is included in all import configurations. If there's no rosdep entry for it IMO it does not need to be included unless not including it would cause interference with the Ubuntu version of the metapackage.

>     * The format `Package, Version | Package, Version` is used in some places, and the format `Package  Package, Version` is used in others. Which one is preferred? The 2nd seems easier to maintain.

We've definitely been refining how we use the filter formula format. `Package | Package, Version` as seen in https://github.com/ros-infrastructure/reprepro-updater/blob/7ec026a809e244431e9b839a5298b1ea17689e08/config/ignition_citadel_gazebo11_ubuntu_focal.yaml#L16-L27 is the currently preferred format.

>     * There's a mismatch of sub-packages between Focal and Buster sometimes. For example, there are 6 packages for `transport10` on Buster and only one on Focal. Should all 6 be added to Focal, or is there some globbing going on?

I think buster is out of date because we were sprinting to get the focal configuration in shape for a deadline and then haven't yet returned to update buster. So at some point we should do that and focal is definitely the recommended template between the two.

chapulina:
Thanks for the feedback, updated Focal in d4ccaff

nuclearsandwich (commented):
The focal config looks good to me.

The buster config is missing some source packages as identified by CI and I think it would ideally be updated to the same `Package | Package, $Version` format as the focal config.

chapulina:
I copied Focal into Buster and CI looks happy now

nuclearsandwich (approved):

nuclearsandwich:
Import jobs 
* https://build.ros.org/job/import_upstream/168/
* https://build.ros2.org/job/import_upstream/326/

PR #110, opened 2021-04-16
Update distribution in version string.

nuclearsandwich:
We might think about dropping the distribution for a glob in the future.


chapulina (approved):
ouch thanks!

nuclearsandwich:
Import jobs 
* https://build.ros.org/job/import_upstream/168/
* https://build.ros2.org/job/import_upstream/326/

PR #112, opened 2021-05-04
Ignition: fix Focal

chapulina:
Focal imports failed after https://github.com/ros-infrastructure/reprepro-updater/pull/111

Compared to the Buster filters, Focal was missing a lot of sub-packages. I had assumed there was some globbing going on, but that doesn't seem to be the case.

This PR makes Focal and Buster as similar as they can be by using `*` for the distro name. This should make it easier to compare them:

```.diff
$ diff -u config/ignition_edifice_debian_buster.yaml config/ignition_edifice_ubuntu_focal.yaml 
--- config/ignition_edifice_debian_buster.yaml	2021-05-04 16:39:32.060849643 -0700
+++ config/ignition_edifice_ubuntu_focal.yaml	2021-05-04 16:39:11.708572984 -0700
@@ -1,8 +1,8 @@
-name: ignition_edifice_debian_buster
-method: http://packages.osrfoundation.org/gazebo/debian-stable
-suites: [buster]
+name: ignition_edifice_ubuntu_focal
+method: http://packages.osrfoundation.org/gazebo/ubuntu-stable
+suites: [focal]
 component: main
-architectures: [amd64, source]
+architectures: [amd64, i386, armhf, arm64, source]
 filter_formula: "\
 (Package (% =ignition-common4), $Version (% 4.0.0-1~*)) |\
 ((Package (% =libignition-common4) |\
```

```.diff
$ diff -u config/ignition_edifice_debian_buster.yaml config/ignition_edifice_ubuntu_focal.yaml 
--- config/ignition_edifice_debian_buster.yaml	2021-05-04 16:39:32.060849643 -0700
+++ config/ignition_edifice_ubuntu_focal.yaml	2021-05-04 16:39:11.708572984 -0700
@@ -1,8 +1,8 @@
-name: ignition_edifice_debian_buster
-method: http://packages.osrfoundation.org/gazebo/debian-stable
-suites: [buster]
+name: ignition_edifice_ubuntu_focal
+method: http://packages.osrfoundation.org/gazebo/ubuntu-stable
+suites: [focal]
 component: main
-architectures: [amd64, source]
+architectures: [amd64, i386, armhf, arm64, source]
 filter_formula: "\
 (Package (% =ignition-common4), $Version (% 4.0.0-1~*)) |\
 ((Package (% =libignition-common4) |\

```

PR #111, opened 2021-04-27
Upgrade Ignition versions

chapulina:
Edifice packages like `ign-msgs7` are failing to install because they require higher versions of packages that are shared with Citadel, like `ign-math6`.

This PR updates all Citadel packages to their latest versions.

PR #113, opened 2021-05-07
Ignition: bump ign-cmake2 and gazebo11 and add ignition-fuel-tools to Buster.

chapulina:
Follow-up to https://github.com/ros-infrastructure/reprepro-updater/pull/112

It's still failing to install `ignition-edifice` on Focal:

```
The following packages have unmet dependencies:
 ignition-edifice : Depends: libignition-cmake2-dev (>= 2.7.0) but 2.6.2-1~focal is to be installed
                    Depends: libignition-common4-dev but it is not going to be installed
                    Depends: libignition-fuel-tools6-dev but it is not going to be installed
                    Depends: libignition-gazebo5-dev but it is not going to be installed
                    Depends: libignition-gui5-dev but it is not going to be installed
                    Depends: libignition-launch4-dev but it is not going to be installed
                    Depends: libignition-math6-dev (>= 6.8.0) but 6.7.0-1~focal is to be installed
                    Depends: libignition-math6-eigen3-dev (>= 6.8.0) but it is not going to be installed
                    Depends: libignition-msgs7-dev but it is not going to be installed
                    Depends: libignition-physics4-dev but it is not going to be installed
                    Depends: libignition-plugin-dev (>= 1.2.0) but 1.1.0-1~focal is to be installed
                    Depends: libignition-rendering5-dev but it is not going to be installed
                    Depends: libignition-sensors5-dev but it is not going to be installed
                    Depends: libignition-tools-dev (>= 1.1.0) but 1.0.0-1~focal is to be installed
                    Depends: libignition-transport10-dev but it is not going to be installed
                    Depends: libignition-utils1-cli-dev but it is not going to be installed
                    Depends: libignition-utils1-dev but it is not going to be installed
                    Depends: libsdformat11-dev but it is not going to be installed
E: Unable to correct problems, you have held broken packages.
```

I narrowed it down to 

```
libignition-utils1-dev : Depends: libignition-cmake2-dev (>= 2.7.0~pre1) but 2.6.2-1~focal is to be installed
```

I think 2.8.0 may have already been out when we the import job was run, which made 2.7 not be pulled and we were stuck with the old 2.6.2.

---

It would be nice to improve this process :upside_down_face:  I actually don't know if just pulling `ign-cmake2` will be enough, so this may not even be the end of this round.

The only suggestion I have is to always pull the latest versions of all libraries (abuse those globs!). I know there's a desire to keep track of what exact versions are being imported, but as noted along this series of PRs (#109 / #110 / #111 / #112), the versions listed on this file aren't accurately reflecting the state of the server.

If we always pull the latest versions:

* We're sure they all work together, because that's actively tested by the Ignition team
* We don't need to edit this file every time we need to make an upgrade (saves time)

### DIFF
--- a/config/ignition_citadel_gazebo11_debian_buster.yaml
+++ b/config/ignition_citadel_gazebo11_debian_buster.yaml
@@ -10,7 +10,7 @@ Package (% =gazebo11-dbg) |\
 Package (% =gazebo11-doc) |\
 Package (% =libgazebo11) |\
 Package (% =libgazebo11-dev)), \
-$Version (% 11.5.0-1~*)) |\
+$Version (% 11.5.1-1~*)) |\
 ((Package (% =ignition-cmake2) |\
 Package (% =libignition-cmake2-dev)) |\
 $Version (% 2.8.0-1~*)) |\
--- a/config/ignition_citadel_gazebo11_debian_buster.yaml
+++ b/config/ignition_citadel_gazebo11_debian_buster.yaml
@@ -13,7 +13,7 @@ Package (% =libgazebo11-dev)), \
 $Version (% 11.5.0-1~*)) |\
 ((Package (% =ignition-cmake2) |\
 Package (% =libignition-cmake2-dev)) |\
-$Version (% 2.7.0-1~*)) |\
+$Version (% 2.8.0-1~*)) |\
 (Package (% =ignition-citadel), $Version (% 1.0.0*))|\
 ((Package (% =ignition-common3) |\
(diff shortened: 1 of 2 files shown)


chapulina:
To make things even more interesting, I see some `ign-cmake2` 2.8.0 packages here, how did they get there? :cry: http://packages.ros.org/ros2/ubuntu/pool/main/i/ignition-cmake2/

nuclearsandwich:
I created #115 to keep the process improvements discussion alive while we address the immediate issues.

Issue #115, opened 2021-05-07
Improve gazebo/igntion import procedures.


nuclearsandwich:
Raised by @chapulina in #113 
It would be nice to improve this process :upside_down_face:  I actually don't know if just pulling `ign-cmake2` will be enough, so this may not even be the end of this round.

The only suggestion I have is to always pull the latest versions of all libraries (abuse those globs!). I know there's a desire to keep track of what exact versions are being imported, but as noted along this series of PRs (#109 / #110 / #111 / #112), the versions listed on this file aren't accurately reflecting the state of the server.

If we always pull the latest versions:

* We're sure they all work together, because that's actively tested by the Ignition team
* We don't need to edit this file every time we need to make an upgrade (saves time)

j-rivero:
The current import approach is not able to check for version constraint dependencies (not in aply CI at least as of today). This can lead to problems like the one raised in https://github.com/gazebosim/gz-cmake/issues/304, that appears when a new version constraint is introduced in an update to a given package for reprepro.

j-rivero:
The CI is not checking if explicit version exists https://github.com/ros-infrastructure/reprepro-updater/pull/166#discussion_r947183877

PR #118, opened 2021-05-07
Update package versions to latest.

nuclearsandwich:
Still trying to settle on installable gazebo and ignition packages. I used some ad-hoc automation to grab the current versions from the Gazebo repositories and check that against what we had configured.

Shell scripts are in Fish so be sure to translate them to their POSIX equivalents

Grab the Sources manifests from the OSRF repostiories, I make the assumption that all binary packages we need to import are present in the configs and that all binary packages matching the source package versions are present in the OSRF repos.

```fish
 curl 'http://packages.osrfoundation.org/gazebo/debian-stable/dists/buster/main/source/Sources.gz' | zcat > debian-buster-sources
 curl 'http://packages.osrfoundation.org/gazebo/ubuntu-stable/dists/focal/main/source/Sources.gz' | zcat > ubuntu-focal-sources
```

I threw a quick Ruby script together just to get this done. `grab-packages.rb`

```ruby
#!/usr/bin/ruby

sources_file = ARGV.shift
sources = File.read(sources_file)
packages_select = ARGV.to_a

Package = Struct.new(:package, :version)
packages = []
package = nil
version = nil
sources.lines.each do |line|
  if line.match /\APackage: (.*)/
    package = $1
  end
  if line.match /\AVersion: (.*)/
    version = $1
  end
  if package and version
    packages << Package.new(package, version)
    package = nil
    version = nil
  end
end

puts packages.select{|p| packages_select.include? p.package}
```

And used it like so to extract package versions for each of the packages in 

```fish
ruby grab-packages ubuntu-focal-sources (cat config/ignition_edifice_ubuntu_focal.yaml | rg '(Package.*Version|\(\(Package)' | cut -d= -f2 | cut -d\) -f1)
ruby grab-packages debian-buster-sources (cat config/ignition_edifice_debian_buster.yaml | rg '(Package.*Version|\(\(Package)' | cut -d= -f2 | cut -d\) -f1)
ruby grab-packages ubuntu-focal-sources (cat config/ignition_citadel_gazebo11_ubuntu_focal.yaml | rg '(Package.*Version|\(\(Package)' | cut -d= -f2 | cut -d\) -f1)
ruby grab-packages debian-buster-sources (cat config/ignition_citadel_gazebo11_debian_buster.yaml | rg '(Package.*Version|\(\(Package)' | cut -d= -f2 | cut -d\) -f1)
```

### DIFF
--- a/config/ignition_citadel_gazebo11_debian_buster.yaml
+++ b/config/ignition_citadel_gazebo11_debian_buster.yaml
@@ -28,7 +28,7 @@ Package (% =libignition-common3-graphics) |\
 Package (% =libignition-common3-graphics-dev) |\
 Package (% =libignition-common3-profiler) |\
 Package (% =libignition-common3-profiler-dev)), \
-$Version (% 3.12.0-1~*)) |\
+$Version (% 3.13.1-1~*)) |\
 ((Package (% =ignition-fuel-tools4) |\
 Package (% =libignition-fuel-tools4) |\
 Package (% =libignition-fuel-tools4-dev)), \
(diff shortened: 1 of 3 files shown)


j-rivero (approved):
updates looks good to me. Thanks for the scripting, we need to find a way of having that in the workflow to avoid problems in the future. Would you mind look a place for them or we can add them to #115 and iterate from there.

PR #120, opened 2021-06-24
Bump catkin_lint to 1.6.14

roehling:
This will also remove Xenial (it's EOL and will no longer be updated in my PPA) and fix a bug in the FilterFormula that prevented propagation of all binary packages (`$Source` instead of `Source` is needed).

### DIFF
--- a/config/catkin_lint.upstream.yaml
+++ b/config/catkin_lint.upstream.yaml
@@ -3,4 +3,4 @@ method: http://ppa.launchpad.net/roehling/ros/ubuntu
 suites: [bionic, focal]
 component: main
 architectures: [i386, amd64, armhf, arm64, source]
-filter_formula: $Source (== catkin-lint), $Version (% 1.6.13-*)
+filter_formula: $Source (== catkin-lint), $Version (% 1.6.14-*)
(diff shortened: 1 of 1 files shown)


config/catkin_lint.upstream.yaml
```diff
@@ -1,6 +1,6 @@
 name: catkin_lint
 method: http://ppa.launchpad.net/roehling/ros/ubuntu
-suites: [xenial, bionic, focal]
+suites: [bionic, focal]
 component: main
 architectures: [i386, amd64, armhf, arm64, source]
-filter_formula: Source (== catkin-lint), $Version (% 1.6.12-*)
+filter_formula: $Source (= catkin-lint), $Version (% 1.6.14-*)
```
nuclearsandwich, line 6:
```suggestion
filter_formula: (Package (= catkin-lint), $Source (= catkin-lint)), $Version (% 1.6.14-*)
```

aptly which is currently in use on the bootstrap repository rather than reprepro doesn't recognize source packages as matching their own $Source so in order to import the catkin-lint source package we also need to match on the package name `catkin-lint`.

We didn't have checks for source package availability previously which is why this didn't come up before.

nuclearsandwich: Whoops. Sorry about e2c9ce3. I didn't transcribe it properly from my local test environment.

nuclearsandwich (approved):

nuclearsandwich:
Import jobs:
* https://build.ros.org/job/import_upstream/180/
* https://build.ros2.org/job/import_upstream/335/

PR #122, opened 2021-07-07
Spike out a change to allow overriding the source package requirement.

nuclearsandwich:
Considering the possibility of using this repository for gurumdds
updates as well but since that's a proprietary package redistributed by
agreement it does not have source packages available and thus we would
need to be able to override our own check.

This is a minimal change which adds a new implicitly boolean field but
I'd like to get feedback on further refactoring to allow overrides to be
package specific (although I have no immediate use case for it).

Examples:

Override all source requirements:

```
override_sources_required: true
```

Override requirements for specific packages

```
override_sources_required:
  - gurumdds-2.6
  - gurumdds-2.7
```

### DIFF
--- a/scripts/aptly/aptly_importer.py
+++ b/scripts/aptly/aptly_importer.py
@@ -175,6 +175,7 @@ class UpdaterConfiguration():
             self.method = self.config['method']
             self.name = self.config['name']
             self.suites = self.config['suites']
+            self.override_sources_required = self.config['override_sources_required']
         except KeyError as e:
             self.__error(f"{e} key was not found in file {input_file}")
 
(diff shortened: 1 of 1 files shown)


scripts/aptly/aptly_importer.py
```diff
@@ -313,13 +317,14 @@ def run(self):
...
             self.__create_aptly_mirror(dist)
-            # 2. Be sure mirror has all source packages
-            self.__log(f'Check all source packages exist')
-            if not self.aptly.exists_all_source_packages(Aptly.ArtifactType.MIRROR,
-                                                         self.__get_mirror_name(dist)):
-                self.__remove_all_generated_mirrors()
-                self.__error(f'{self.__get_mirror_name(dist)} does not have a source package. Removing generated mirrors')
-            self.__log_ok('All source packages exist in the mirror')
+            # 2. Be sure mirror has all source packages unless the requirement is overridden.
+            if not self.config.override_sources_required:
```
tfoote, line 321: It always logs that it's verifying the sources. For this case it would be good to log that it's explicitly skipping the check based on the config.

j-rivero:
depending on urgency, it would be great to add a test for this case. Otherwise, looks good to me.

nuclearsandwich:
I agree adding a test is a requirement for getting this in and I like Tully's suggestion of logging the override.

PR #128, opened 2021-10-11
Add Ignition Fortress

chapulina:
Fortress is Ignition's latest LTS and will be the official version to use with Humble, so I'd like to start using Fortress on Rolling soon.

I want to get some more releases in before importing the packages. Opening the PR for visibility.

Part of https://github.com/ignition-tooling/release-tools/issues/459


config/ignition_fortress_debian_buster.yaml
```diff
@@ -0,0 +1,96 @@
+name: ignition_fortress_debian_buster
+method: http://packages.osrfoundation.org/gazebo/debian-stable
+suites: [buster]
```
nuclearsandwich, line 3: Rolling is going to switch to Debian Bullseye alongside Ubuntu Jammy. Are we currently building Fortress for Bullseye?

config/ignition_fortress_ubuntu_focal.yaml
```diff
@@ -0,0 +1,96 @@
+name: ignition_fortress_ubuntu_focal
+method: http://packages.osrfoundation.org/gazebo/ubuntu-stable
+suites: [focal]
```
nuclearsandwich, line 3: I think we're hoping for Fortress to be upstream in Ubuntu Jammy correct? We'll have these Fortress packages for Focal but over the next month Rolling development will jump over to the Jammy pre-releases and our upstreaming efforts are still underway. I can see some sense in adding Fortress to the bootstrap repo for Jammy in the interim and then removing those packages once they become available upstream. But I have to refresh myself on how hard it is to _remove_ packages from the bootstrap repository and ROS repositories once imported.

nuclearsandwich (commented):
Some comments about the target distributions since Rolling is changing in the next 30-60 days but otherwise looks well formed.

config/ignition_fortress_debian_buster.yaml
chapulina: Not yet, see https://github.com/ignition-tooling/release-tools/issues/566

config/ignition_fortress_ubuntu_focal.yaml
chapulina: Correct. One thing to keep in mind though is that relying on upstream will lock us into a fixed version, which wouldn't be great.

I know this has been brought up before, but I'm not sure we got to a consensus. I think ideally we'd have a way to continuously push updates to the Ignition packages consumed by ROS users.

chapulina:
> Rolling is changing in the next 30-60 days

It would be really nice to get Fortress into Rolling before that. Can we add it now while we're still on Focal and address Jammy when the time comes?

nuclearsandwich:
> It would be really nice to get Fortress into Rolling before that. Can we add it now while we're still on Focal and address Jammy when the time comes?

That works for me, I just wanted to make sure that you knew the outlook and wanted to spend the effort still.

chapulina:
> That works for me, I just wanted to make sure that you knew the outlook and wanted to spend the effort still.

Thanks. We'll deal with Jammy / Bullseye soon. Meanwhile, I think this PR is ready.

j-rivero:
Import to build.ros.org here [Build Status](https://build.ros.org/job/import_upstream/200/)

j-rivero:
ROS2 upload [Build Status](https://build.ros2.org/job/import_upstream/361/)

PR #130, opened 2021-11-13
Add bullseye key to test scripts.

nuclearsandwich:
This is pretty brittle in general, I wonder if we can trust a top level
key or import all Debian keys as we do with the Ubuntu keys (which is
easy because we're running the CI on Ubuntu.)


chapulina (approved):
Ouch yeah pretty brittle, but it fixes CI for now :+1:

PR #129, opened 2021-11-12
Update Gazebo 11 and Ignition Citadel

chapulina:
Periodic update to the latest versions.

I'd like to get this in before #128 , to reduce that PR

PR #131, opened 2021-11-16
Upgrade Ignition Edifice packages

chapulina:
Similar to #129, periodic update to the latest versions.

I'd like to get this in before #128 , to reduce that PR

One thing I did differently here was to remove the revision `-1` from the versions, because some packages were released as `-2`

### DIFF
--- a/config/ignition_edifice_debian_buster.yaml
+++ b/config/ignition_edifice_debian_buster.yaml
@@ -16,30 +16,30 @@ Package (% =libignition-common4-graphics) |\
 Package (% =libignition-common4-graphics-dev) |\
 Package (% =libignition-common4-profiler) |\
 Package (% =libignition-common4-profiler-dev)), \
-$Version (% 4.4.0*)) |\
+$Version (% 4.4.0-*)) |\
 (Package (% =ignition-edifice), $Version (% 1.0.2*)) |\
 ((Package (% =ignition-fuel-tools6) |\
 Package (% =libignition-fuel-tools6) |\
 Package (% =libignition-fuel-tools6-dev)), \
-$Version (% 6.1.0*)) |\
+$Version (% 6.1.0-*)) |\
 ((Package (% =ignition-gazebo5) |\
 Package (% =libignition-gazebo5) |\
 Package (% =libignition-gazebo5-dev) |\
 Package (% =libignition-gazebo5-plugins)), \
-$Version (% 5.3.0*)) |\
+$Version (% 5.3.0-*)) |\
 ((Package (% =ignition-gui5) |\
 Package (% libignition-gui5) |\
 Package (% libignition-gui5-dev)), \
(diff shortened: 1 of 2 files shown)


nuclearsandwich:
> One thing I did differently here was to remove the revision `-1` from the versions, because some packages were released as `-2`

In general I'm very much a fan of fully specified releases since there's really no better expression of what we _want_ in the bootstrap repository than these configs. But since we don't have any enforcement that every clause in a filter expression actually matches a package I can't really justify being dogmatic about it. I think it does make sense to include the `-` separator literal to prevent a version such as `1.4.1*` from matching `1.4.11` in addition to the intended `1.4.1-2`.

nuclearsandwich (changes requested):
Even if we don't include the debinc value in the fixed portion of the version glob I think it's best to include the debinc separator character `-` to prevent unintended matches.

chapulina:
Makes sense, added the `-` back in 35e4b31

config/ignition_edifice_debian_buster.yaml
```diff
@@ -16,30 +16,30 @@ Package (% =libignition-common4-graphics) |\
 Package (% =libignition-common4-graphics-dev) |\
 Package (% =libignition-common4-profiler) |\
 Package (% =libignition-common4-profiler-dev)), \
-$Version (% 4.0.0-1~*)) |\
-(Package (% =ignition-edifice), $Version (% 1.0.0-1~*)) |\
+$Version (% 4.4.0-*)) |\
+(Package (% =ignition-edifice), $Version (% 1.0.2*)) |\
```
nuclearsandwich, line 20: Looks like one was missed.

```suggestion
(Package (% =ignition-edifice), $Version (% 1.0.2-*)) |\
```

config/ignition_edifice_ubuntu_focal.yaml
```diff
@@ -16,30 +16,30 @@ Package (% =libignition-common4-graphics) |\
 Package (% =libignition-common4-graphics-dev) |\
 Package (% =libignition-common4-profiler) |\
 Package (% =libignition-common4-profiler-dev)), \
-$Version (% 4.0.0-1~*)) |\
-(Package (% =ignition-edifice), $Version (% 1.0.0-1~*)) |\
+$Version (% 4.4.0-*)) |\
+(Package (% =ignition-edifice), $Version (% 1.0.2*)) |\
```
nuclearsandwich, line 20:
```suggestion
(Package (% =ignition-edifice), $Version (% 1.0.2-*)) |\
```

nuclearsandwich (approved):
There was one "missed" edit common to both files but other than those it looks good and passes CI.

nuclearsandwich:
Import jobs:
https://build.ros.org/job/import_upstream/191/
https://build.ros2.org/job/import_upstream/346/

PR #136, opened 2021-12-23
Update 6 recently released colcon packages

cottsay:
Packages pushed to: https://packagecloud.io/dirk-thomas/colcon


nuclearsandwich:
Could we get f19b710 into its own PR so that it doesn't get squashed out of the history?

cottsay:
> Could we get [f19b710](https://github.com/ros-infrastructure/reprepro-updater/commit/f19b71051971d99efdd57fa1c0846d50e1758ca6) into its own PR so that it doesn't get squashed out of the history?

Done: #137

cottsay:
...and rebased, now that #137 is merged.

nuclearsandwich (approved):

cottsay:
Import is done:
* https://build.ros2.org/job/import_upstream/363/
* https://build.ros.org/job/import_upstream/204/

PR #137, opened 2021-12-23
Target Jammy in place of EOL Groovy in scripts

cottsay:
It appears that groovy is no longer on archive.ubuntu.com.


chapulina (approved):
Seems reasonable to me

j-rivero (approved):

j-rivero:
Thanks @cottsay !

PR #139, opened 2021-12-23
Update to colcon-core 0.7.1

cottsay:
Published to: https://packagecloud.io/dirk-thomas/colcon


cottsay:
Import jobs:
https://build.ros2.org/job/import_upstream/365/
https://build.ros.org/job/import_upstream/205/

PR #140, opened 2021-12-26
Bump colcon-ed to 0.2.1

Briancbn:
Signed-off-by: Chen Bainian <chenbn@artc.a-star.edu.sg>

Published in https://packagecloud.io/dirk-thomas/colcon. Focal and Bionic is also in there, published 2 weeks ago.


cottsay:
Something went wrong pulling this package in:
```
ERROR: unable to publish: unable to process packages: error linking file to /home/apt/public/repos/ros_bootstrap/pool/main/p/python3-colcon-ed/python3-colcon-ed_0.2.1-1.dsc: file already exists and is different
```

It looks like it was able to handle artful, bionic, and bullseye properly before hitting the error.

@nuclearsandwich, please give me a hand with this when you have some time.

nuclearsandwich:
This problem arises when different suites have different copies of the same (by name and version) package. It can occur if the Debian and Ubuntu packages are built separately with matching versions but different copies of an upstream tarball because the ROS repositories use a single apt layout for Debian and Ubuntu repositories.

In this case it looks like the issue is that there is a copy of this package in the [Xenial](https://packagecloud.io/dirk-thomas/colcon/packages/ubuntu/xenial/python3-colcon-ed_0.2.1-1.dsc) and [Focal](https://packagecloud.io/dirk-thomas/colcon/packages/ubuntu/focal/python3-colcon-ed_0.2.1-1.dsc) repositories which does not match the one uploaded two days ago for [Jammy](https://packagecloud.io/dirk-thomas/colcon/packages/ubuntu/focal/python3-colcon-ed_0.2.1-1.dsc), [Stretch](https://packagecloud.io/dirk-thomas/colcon/packages/debian/stretch/python3-colcon-ed_0.2.1-1.dsc), [Buster](https://packagecloud.io/dirk-thomas/colcon/packages/debian/buster/python3-colcon-ed_0.2.1-1.dsc), and [Bullseye](https://packagecloud.io/dirk-thomas/colcon/packages/debian/bullseye/python3-colcon-ed_0.2.1-1.dsc)

I pulled a few of them locally to confirm.

```
sha256sum *.python3-colcon-ed_0.2.1-1.dsc
639a5dcc4bf12af51d9b55340fe297b80bb828daee392b3444f863729533960a  bullseye.python3-colcon-ed_0.2.1-1.dsc
639a5dcc4bf12af51d9b55340fe297b80bb828daee392b3444f863729533960a  buster.python3-colcon-ed_0.2.1-1.dsc
97890a07c81a8571efa4d218041132fd647db78d667ef6844a73b0ede4d39a3b  focal.python3-colcon-ed_0.2.1-1.dsc
639a5dcc4bf12af51d9b55340fe297b80bb828daee392b3444f863729533960a  jammy.python3-colcon-ed_0.2.1-1.dsc
```

We'll need to do two things to recover.

1. Remove colcon-ed 0.2.1-1 from all repositories. We could be more precise here but it's easiest for consistencies sake to remove it from all of them. When we do so the previous version of the package in each repository will still be present so this won't remove colcon-ed entirely, even temporarily.
2. Publish a package revision of colcon-ed 0.2.1 to all target suites on PackageCloud so that they're all using the same artifact.

Then we can re-run the import fetching the latest and consistent version of colcon-ed.

I couldn't see via a quick grep a way to pass the stdeb `--debian-version` command line arg through from publish-python but an additional config stanza can be temporarily added to stdeb.cfg
```
Debian-Version: 2
```

When running this release be sure not to push to pypi since that will error when trying to publish a version that has already been published. The nuclear I-don't-want-to-think-about-it option is to bump and tag a 0.2.2 and then run the release again with all suites.

Briancbn:
Sorry about the mess here. I think this is because of the following

> Focal and Bionic is also in there, published 2 weeks ago.

I rebuilt the deb & dsc, but not for Focal and Bionic. Do you want me to release a 0.2.2 instead, this time correctly, or wait for you guys to resolve it in 0.2.1, as files need to be removed on the server side?

> I couldn't see via a quick grep a way to pass the stdeb `--debian-version` command line arg through from publish-python but an additional config stanza can be temporarily added to stdeb.cfg
> 
> ```
> Debian-Version: 2
> ```

This is also a good option too, but I am not sure if publish_python will upload it correctly, and I am a little bit scared to try. Please let me know your thoughts.

cottsay:
> Sorry about the mess here.

No worries, this isn't your fault.

> Do you want me to release a 0.2.2...?

That would probably be the fastest way to solve the problem.

PR #143, opened 2022-02-15
Add new import configuration for Ignition Fortress on Ubuntu Jammy.

nuclearsandwich:
Currently only ignition-cmake and ignition-math are available.

I'm prototyping a script in [another branch](https://github.com/ros-infrastructure/reprepro-updater/tree/nuclearsandwich/generating-ignition-configs) to generate these config files by inspecting the upstream repository hence the slightly different formatting which hopefully improves readability.

### DIFF
--- a/config/ignition_fortress_ubuntu_jammy.yaml
+++ b/config/ignition_fortress_ubuntu_jammy.yaml
@@ -1,0 +1,17 @@
+name: ignition_fortress_ubuntu_jammy
+method: http://packages.osrfoundation.org/gazebo/ubuntu-stable
+suites: [jammy]
+component: main
+architectures: [amd64, i386, armhf, arm64, source]
+filter_formula: "\
+((Package (= ignition-cmake2) |\
+ (Package (= libignition-cmake2-dev)
+), $Version (% 2.10.0-*) |\
+((Package (= ignition-math6) |\
+ (Package (= libignition-math6) |\
+ (Package (= libignition-math6-dbg) |\
+ (Package (= libignition-math6-dev) |\
+ (Package (= libignition-math6-eigen3-dev) |\
+ (Package (= python3-ignition-math6)
+), $Version (% 6.10.0-*) \
+"


config/ignition_fortress_ubuntu_jammy.yaml
```diff
@@ -0,0 +1,17 @@
+name: ignition_fortress_ubuntu_jammy
+method: http://packages.osrfoundation.org/gazebo/ubuntu-stable
+suites: [jammy]
+component: main
+architectures: [amd64, i386, armhf, arm64, source]
```
j-rivero, line 5:
```suggestion
architectures: [amd64, armhf, arm64, source]
```
Jammy does not support i386

config/ignition_fortress_ubuntu_jammy.yaml
```diff
@@ -0,0 +1,17 @@
...
+filter_formula: "\
+(Package (= ignition-cmake2) |\
+ Package (= libignition-cmake2-dev)
+), $Version (% 2.10.0-*) |\
+(Package (= ignition-math6) |\
+ Package (= libignition-math6) |\
+ Package (= libignition-math6-dbg) |\
+ Package (= libignition-math6-dev) |\
+ Package (= libignition-math6-eigen3-dev) |\
+ Package (= python3-ignition-math6)
```
j-rivero, line 15:
```suggestion
 Package (= python3-ignition-math6) |\
 Package (= ruby-ignition-math6)
```
We now have a nice ruby bindings in ign-math6

j-rivero (changes requested):
A couple of important fixes. Given that we have in Jammy the same exact versions, we might find problems at the time of installation since some of them would resolve to the packages in Ubuntu. I think that I can workaround about this problem by adding the osrf suffix to version.

For example, in ignition-math we have the ds suffix in Debian (dfsg is also common) but our osrf suffix would take precedence over them and over the vanilla versions:

```bash
❯ dpkg --compare-versions 6.10.0+osrf1-1 gt 6.10.0+ds2-1 && echo "yes"
yes
❯ dpkg --compare-versions 6.10.0+osrf1-1 gt 6.10.0-1 && echo "yes"
yes
❯ dpkg --compare-versions 6.10.0+osrf1-1 gt 6.10.0+dfsg-1 && echo "yes"
yes
```
I would need to re-release the packages, should not take too long.

j-rivero:
> For example, in ignition-math we have the ds suffix in Debian (dfsg is also common) but our osrf suffix would take precedence over them and over the vanilla versions:

Thinking twice this would require some changes in the building scripts since the orig tarball would not be the same that the version (6.10.0 vs 6.10.0+osrf1 in the example above). Another easy way would be to increase the release version to something like 99 or similar. This does not require of any change.

nuclearsandwich:
> Thinking twice this would require some changes in the building scripts since the orig tarball would not be the same that the version (6.10.0 vs 6.10.0+osrf1 in the example above). Another easy way would be to increase the release version to something like 99 or similar. This does not require of any change.

We use the super-high release increment trick for other ros-infrastructure packages so that would probably be my preference. It means we have to tweak the import configuration to include the release increment but that is very doable.

j-rivero:
All right, we should have the `-99` versions of cmake2 and math6 on Jammy for amd63, armhf and arm64. We don't need more changes to the PR I think.

nuclearsandwich:
> All right, we should have the `-99` versions of cmake2 and math6 on Jammy for amd63, armhf and arm64. We don't need more changes to the PR I think.

Thanks! The script found the updated ignition-math6 but not ignition-cmake2.

j-rivero (approved):

nuclearsandwich:
Import upstream jobs:
* https://build.ros.org/job/import_upstream/214/ (no-op for completeness since there is no Jammy repository for the ROS 1 farm)
* https://build.ros2.org/job/import_upstream/375/

PR #144, opened 2022-02-16
Add a script for generating import configurations from the live Ignition repositories.

nuclearsandwich:
:construction: :construction: :construction: :construction: :construction: :construction: :construction: :construction: :construction: :construction: 

This is a work-in-progress!

:construction: :construction: :construction: :construction: :construction: :construction: :construction: :construction: :construction: :construction: 

This script reads data from the packages.osrfoundation.org repository and prints out a reprepro-updater import configuration.

*Caveats and addendums*

* Right now it only supports the `-stable` repositories, and has a set of packages that essentially hard-codes Ignition Fortress despite taking the `--ignition-suite` as an argument. 
* Another limitation is that it looks _only_ at the binary-amd64 package information so it implicitly assumes no disparities between architectures.
* The parsing of the Debian `Packages` file is extremely rudimentary but presently sufficient in order to avoid dependencies outside the Python standard library.

*But that's enough with the caveats, what cool things does it do?*

With a builtin list of igniton suite source packages, it will find group up all of the binary packages generated by that source and create a valid reprepro import configuration for use in the ros_bootstrap repository. This will make it much easier to keep the bootstrap repository in sync with the current releases from the packages.osrfoundation.org repository and could eventually be run out-of-tree as part of the upstream team's release automation.

#143 was created using early prototypes of this script and I have another branch which reformats the existing ignition configurations to conform to the same style so that I can use the diff between the manually created configs and the ones this script generates to refine and improve the script, and especially, add support for other Ignition suites.

### DIFF
--- a/scripts/generate_ignition_import_config.py
+++ b/scripts/generate_ignition_import_config.py
@@ -1,7 +1,8 @@
 #!/usr/bin/env python3
 
-import urllib.request
+import argparse
 import sys
+import urllib.request
 
 from dataclasses import dataclass
 from typing import List, Union
@@ -41,10 +42,14 @@ class PackageGroup:
     packages: None|List[str]
     version_spec: str|None
 
-IGNITION_SUITE='fortress'
+parser = argparse.ArgumentParser(description='Generate import configurations for ignition packages.')
+parser.add_argument('--os', type=str)
+parser.add_argument('--suite', type=str)
+parser.add_argument('--ignition-suite', type=str)
+args = parser.parse_args(sys.argv[1:])
 
-PACKAGES = [
-        PackageGroup(f'ignition-{IGNITION_SUITE}', packages=None, version_spec=None),
+package_groups = [
+        PackageGroup(f'ignition-{args.ignition_suite}', packages=None, version_spec=None),
         PackageGroup('ignition-cmake2', packages=None, version_spec=None),
         PackageGroup('ignition-fuel-tools7', packages=None, version_spec=None),
         PackageGroup('ignition-gazebo6', packages=None, version_spec=None),
@@ -60,15 +65,12 @@ PACKAGES = [
         PackageGroup('sdformat12', packages=None, version_spec=None),
     ]
 
-OS = 'ubuntu'
-TARGET_REPO = f'http://packages.osrfoundation.org/gazebo/{OS}-stable'
-DISTS = ('focal', 'jammy', 'bullseye', 'buster')
-DISTRO = DISTS[1]
+target_repo = f'http://packages.osrfoundation.org/gazebo/{args.os}-stable'
 
-resp = urllib.request.urlopen(f'{TARGET_REPO}/dists/{DISTRO}/main/binary-amd64/Packages')
+resp = urllib.request.urlopen(f'{target_repo}/dists/{args.suite}/main/binary-amd64/Packages')
 packages_file = PackagesFile(resp.read().decode())
 
-for group in PACKAGES:
+for group in package_groups:
     expected_group_version = None
     group.packages = [pkg.name for pkg in packages_file.packages.values() if pkg.source_package == group.source_package]
     for p in group.packages:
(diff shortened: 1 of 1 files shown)


nuclearsandwich:
No reviews yet but FYI @j-rivero @chapulina

chapulina:
> Right now it only supports the -stable repositories

This should be enough for our needs :+1:

nuclearsandwich:
@j-rivero I only just learned about https://github.com/ignition-tooling/gazebodistro/pull/4 today and it looks like we thought very similarly here.

What I like from the script in that repository is that it uses the gazebodistro repos files at runtime to get the base package information, and then includes all dependencies of selected packages from the packages.osrfoundation.org repository.

What I like about this script is that it generates the exact configuration file reprepro-updater wants to consume and it can be run on any host platform rather than being dependent on the system repositories for state.

I think we should talk through the current status of each script and see which one to concentrate our efforts on.

j-rivero:
I'm giving a try to the tool and can not make it work with python 3.8 and 3.9.5:
```
❯ python3.9 generate_ignition_import_config.py 
Traceback (most recent call last):
  File "/home/jrivero/code/infra/reprepro-updater/scripts/generate_ignition_import_config.py", line 44, in <module>
    class PackageGroup:
  File "/home/jrivero/code/infra/reprepro-updater/scripts/generate_ignition_import_config.py", line 46, in PackageGroup
    packages: None|List[str] = None
TypeError: unsupported operand type(s) for |: 'NoneType' and '_GenericAlias'
```
Is this kind o [hinting support](https://peps.python.org/pep-0585/) in newer versions or there is a bug somehow?

> @j-rivero I only just learned about [gazebo-tooling/gazebodistro#4](https://github.com/gazebo-tooling/gazebodistro/pull/4) today and it looks like we thought very similarly here.
> 
> What I like from the script in that repository is that it uses the gazebodistro repos files at runtime to get the base package information, and then includes all dependencies of selected packages from the packages.osrfoundation.org repository.
> 
> What I like about this script is that it generates the exact configuration file reprepro-updater wants to consume and it can be run on any host platform rather than being dependent on the system repositories for state.
> 
> I think we should talk through the current status of each script and see which one to concentrate our efforts on.

+1 Did not really like the gazebodistro PR so +1 to merge efforts here.

scripts/generate_ignition_import_config.py
```diff
@@ -0,0 +1,167 @@
...
+
+version_spec_re = re.compile('(?P<version>[^-]+)-(?P<inc>\d+)(?P<rest>.*)')
+for group in package_groups:
+    expected_group_version = None
+    group.packages = [pkg.name for pkg in packages_file.packages.values() if pkg.source_package == group.source_package and pkg.name not in group.skip_packages]
+    for p in group.packages:
+        if not expected_group_version:
+            expected_group_version = packages_file.packages[p].version
+        elif packages_file.packages[p].version != expected_group_version:
+            print(f'{p} in {group.packages[0]} does not match expected version {expected_group_version}', file=sys.stderr)
```
j-rivero, line 133: The check is nice to have. However It can detect problems with versions for orphaned packages, like the bad named `libgz-math6-eigen-dev` now renamed to `libgz-math6-eigen3-dev`. If we keep the exit after the check, the script can not continue.

My proposal is to remove the exit and leave the warning goes into stderr which does not affect standard redirections and diffs.

scripts/generate_ignition_import_config.py
```diff
@@ -0,0 +1,167 @@
...
+            exit(2)
```
j-rivero, line 134:
```suggestion
            group.packages.remove(p)
```
The check is nice to have. However It can detect problems with versions for orphaned packages, like the bad named `libgz-math6-eigen-dev` now renamed to `libgz-math6-eigen3-dev`. If we keep the exit after the check, the script can not continue.

A possible path forward would be to leave the warning but remove the package from the expected imports.

azeey:
@azeey

PR #145, opened 2022-04-08
Upgrade Edifice packages

chapulina:
The last sync for Edifice, since it's now EOL.

* Similar to #131 
* Part of https://github.com/ignition-tooling/release-tools/issues/655

Note that I've also bumped `ign-cmake2` and `ign-math6`, which are still supported under Citadel, but are Edifice dependencies as well.


nuclearsandwich:
superseded by https://github.com/ros-infrastructure/reprepro-updater/pull/146

PR #146, opened 2022-04-09
Update Ignition Edifice packages using generator script.

nuclearsandwich:
This supersedes #145 which updates version numbers but does not actually match the current packages in the packages.osrfoundation.org repository.

The first commit on this PR, a3e52e855b395165396950a71180577a384e7d35 updates the format of the config file while not changing any contents so that it's easier to diff between the updated, generated content, and the previous content.

I recommend that reviewers use custom compare links to review the changes without formatting differences.
* [ubuntu_focal](https://github.com/ros-infrastructure/reprepro-updater/compare/dcacc2424401911524a379ed909b328d8f5755c5..nuclearsandwich/edifice-updates-2022-04-05#diff-9f768fbe2a06da79aae35728f17b7c509b43ed450effff17065caae1d7956a74)
* [debian_buster](https://github.com/ros-infrastructure/reprepro-updater/compare/d6c0870763c6555b9e9490a74a2733aeeda89876..nuclearsandwich/edifice-updates-2022-04-05#diff-fb26f866f50681cde8fb732babb3558d5b5a5e8d105616ec22593d05a92d8bb1)

### DIFF
--- a/config/ignition_edifice_ubuntu_focal.yaml
+++ b/config/ignition_edifice_ubuntu_focal.yaml
@@ -5,7 +5,7 @@ component: main
 architectures: [amd64, armhf, arm64, source]
 filter_formula: "\
 (Package (= ignition-edifice)
-), $Version (% 1.0.2-*) |\
+), $Version (% 1.0.3-1*) |\
 (Package (= ignition-common4) |\
  Package (= libignition-common4) |\
  Package (= libignition-common4-av) |\
@@ -18,33 +18,33 @@ filter_formula: "\
  Package (= libignition-common4-graphics-dev) |\
  Package (= libignition-common4-profiler) |\
  Package (= libignition-common4-profiler-dev)
-), $Version (% 4.4.0-1*) |\
+), $Version (% 4.5.0-1*) |\
 (Package (= ignition-fuel-tools6) |\
  Package (= libignition-fuel-tools6) |\
  Package (= libignition-fuel-tools6-dev)
-), $Version (% 6.1.0-*) |\
+), $Version (% 6.2.0-1*) |\
 (Package (= ignition-gazebo5) |\
  Package (= libignition-gazebo5) |\
+ Package (= libignition-gazebo5-dbg) |\
  Package (= libignition-gazebo5-dev) |\
  Package (= libignition-gazebo5-plugins)
-), $Version (% 5.3.0-*) |\
+), $Version (% 5.4.0-1*) |\
 (Package (= ignition-gui5) |\
  Package (= libignition-gui5) |\
  Package (= libignition-gui5-dev)
-), $Version (% 5.3.0-*) |\
+), $Version (% 5.5.0-1*) |\
 (Package (= ignition-launch4) |\
  Package (= libignition-launch4) |\
  Package (= libignition-launch4-dev)
-), $Version (% 4.1.0-*) |\
+), $Version (% 4.1.0-1*) |\
 (Package (= ignition-msgs7) |\
  Package (= libignition-msgs7) |\
- Package (= libignition-msgs7-dev) |\
- Package (= libignition-msgs7-dbg)
-), $Version (% 7.2.0-*) |\
+ Package (= libignition-msgs7-dev)
+), $Version (% 7.3.0-1*) |\
 (Package (= ignition-physics4) |\
  Package (= libignition-physics4) |\
- Package (= libignition-physics4-bullet-dev) |\
  Package (= libignition-physics4-bullet) |\
+ Package (= libignition-physics4-bullet-dev) |\
  Package (= libignition-physics4-core-dev) |\
  Package (= libignition-physics4-dartsim) |\
  Package (= libignition-physics4-dartsim-dev) |\
@@ -56,7 +56,7 @@ filter_formula: "\
  Package (= libignition-physics4-tpe-dev) |\
  Package (= libignition-physics4-tpelib) |\
  Package (= libignition-physics4-tpelib-dev)
-), $Version (% 4.3.0-*) |\
+), $Version (% 4.3.0-1*) |\
 (Package (= ignition-rendering5) |\
  Package (= libignition-rendering5) |\
  Package (= libignition-rendering5-core-dev) |\
@@ -65,7 +65,7 @@ filter_formula: "\
  Package (= libignition-rendering5-ogre1-dev) |\
  Package (= libignition-rendering5-ogre2) |\
  Package (= libignition-rendering5-ogre2-dev)
-), $Version (% 5.2.0-*) |\
+), $Version (% 5.2.1-1*) |\
 (Package (= ignition-sensors5) |\
  Package (= libignition-sensors5) |\
  Package (= libignition-sensors5-air-pressure) |\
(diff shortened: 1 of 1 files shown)


nuclearsandwich:
Currently includes a draft of https://github.com/ros-infrastructure/reprepro-updater/pull/147 cherry-picked. I want to rebase that out once #147 is merged.

config/ignition_edifice_ubuntu_focal.yaml
```diff
@@ -2,119 +2,126 @@ name: ignition_edifice_ubuntu_focal
...
+ Package (= libignition-launch4-dev)
+), $Version (% 4.1.0-1*) |\
+(Package (= ignition-math6) |\
+ Package (= libignition-math6) |\
+ Package (= libignition-math6-dbg) |\
+ Package (= libignition-math6-dev) |\
+ Package (= libignition-math6-eigen3-dev) |\
+ Package (= python3-ignition-math6) |\
+ Package (= ruby-ignition-math6)
+), $Version (% 6.10.0-1*) |\
```
chapulina, line 47: `ign-math6` is also on the Gazebo 11 / Citadel config, is it ok to have duplicates?

chapulina (commented):
The version numbers look good to me.

We also need to upgrade `ign-cmake2` and `ign-math6`, which are common between Citadel and Edifice, see https://github.com/ros-infrastructure/reprepro-updater/pull/145/files#diff-e5e0a0f10c0f27937b901ac927c2c6b40354686df642be49ec8c2811db37dbd1. I see you added `math6` here. Should it be moved to the Citadel file or should `ign-cmake2` be duplicated here?

config/ignition_edifice_ubuntu_focal.yaml
nuclearsandwich: I've probably swung my opinion on this from what it used to be.
When we were doing this by hand, it made sense (to me) to have each config cover disjoint sets of packages since otherwise you could end up in a situation where packages were out of "sync" between sets. (For example: the Citadel config could have math 6.9.0 while the Edifice config could have 6.10.0) but I think with the extra scripting to facilitate uploads it will make more sense to update everything at once and the chances for that mis-synchronization go down. So I think it is best to have every package needed by Edifice in the Edifice config and so on for the others.

nuclearsandwich: I updated the PR. Now https://github.com/ros-infrastructure/reprepro-updater/pull/146/commits/9b21a9f811266df3cf125e93a20d5bb0efa60233 _just_ bumps the versions and cmake and math are added in https://github.com/ros-infrastructure/reprepro-updater/pull/146/commits/5aa4818feedbd6c70dc0a6bc74b92edf3f06151f

config/ignition_edifice_debian_buster.yaml
```diff
@@ -2,119 +2,148 @@ name: ignition_edifice_debian_buster
 method: http://packages.osrfoundation.org/gazebo/debian-stable
 suites: [buster]
 component: main
-architectures: [amd64, source]
+architectures: [amd64, armhf, arm64, source]
```
nuclearsandwich, line 5: I would like to just standardize on pulling these architectures. Although I could be convinced to give up hope for armhf ever coming back.

chapulina: From time to time Jose tries to fix them, but it's not a priority for us, so I'd be ok removing them if they cause headaches.

config/ignition_edifice_debian_buster.yaml
```diff
@@ -2,120 +2,148 @@ name: ignition_edifice_debian_buster
...
 filter_formula: "\
-(Package (= ignition-edifice)
-), $Version (% 1.0.2-*) |\
+(Package (= ignition-edifice) \
+), $Version (% 1.0.0-1*) |\
```
nuclearsandwich, line 8: It's odd that the current repository appears to have regressed in version compared to what was configured previously.

Looking at the ROS 2 repository: http://repo.ros2.org/ubuntu/main/pool/main/i/ignition-edifice/

It does not appear that ignition-edifice on Buster ever had a successful 1.0.2 import since the import job will not downgrade a package.
```
ignition-edifice_1.0.0-1~buster.debian.tar.xz      16-Apr-2021 13:57                1300
ignition-edifice_1.0.0-1~buster.dsc                16-Apr-2021 13:57                1582
ignition-edifice_1.0.0-1~buster_amd64.deb          16-Apr-2021 13:57                2084
```

config/ignition_edifice_debian_buster.yaml
```diff
@@ -2,120 +2,148 @@ name: ignition_edifice_debian_buster
...
+ Package (= libignition-fuel-tools6-dev) \
+), $Version (% 6.2.0-1*) |\
 (Package (= ignition-gazebo5) |\
  Package (= libignition-gazebo5) |\
+ Package (= libignition-gazebo5-dbg) |\
  Package (= libignition-gazebo5-dev) |\
- Package (= libignition-gazebo5-plugins)
-), $Version (% 5.3.0-*) |\
+ Package (= libignition-gazebo5-plugins) \
+), $Version (% 5.2.0-1*) |\
```
nuclearsandwich, line 37: Here's another "downgrade" but in this case the imported package is actually still at 5.0.0 so this will affect an upgrade.

```
ignition-gazebo5_5.0.0-1~buster.debian.tar.xz      16-Apr-2021 13:57                2068
ignition-gazebo5_5.0.0-1~buster.dsc                16-Apr-2021 13:57                1858
libignition-gazebo5-dev_5.0.0-1~buster_amd64.deb   16-Apr-2021 13:57              137996
libignition-gazebo5-plugins_5.0.0-1~buster_amd6..> 16-Apr-2021 13:57             3013128
libignition-gazebo5_5.0.0-1~buster_amd64.deb       16-Apr-2021 13:57             3906496
```

config/ignition_edifice_debian_buster.yaml
```diff
@@ -2,120 +2,148 @@ name: ignition_edifice_debian_buster
...
  Package (= libignition-physics4-sdf-dev) |\
  Package (= libignition-physics4-tpe) |\
+ Package (= libignition-physics4-tpe-dbgsym) |\
  Package (= libignition-physics4-tpe-dev) |\
  Package (= libignition-physics4-tpelib) |\
- Package (= libignition-physics4-tpelib-dev)
-), $Version (% 4.3.0-*) |\
+ Package (= libignition-physics4-tpelib-dbgsym) |\
+ Package (= libignition-physics4-tpelib-dev) \
+), $Version (% 4.2.0-2*) |\
```
nuclearsandwich, line 73: Previously configured: 4.3.0
Actually available in the repository: 4.2.0

Currently imported in ROS: 4.1.0

config/ignition_edifice_ubuntu_focal.yaml
chapulina: I'm ok with whatever works for you. I was just checking.

config/ignition_edifice_debian_buster.yaml
chapulina: I'm not surprised, when Buster builds fail we usually take a long time to go back and fix them.

> It does not appear that ignition-edifice on Buster ever had a successful 1.0.2 import 

That's one of the reasons I say that these yaml files don't really correspond to the state of the packages. Imports have been failing silently since we started this process 😬

nuclearsandwich: The entire purpose of the exercise is to get _closer_.
If I could find a non-exhaustive way to do it I'd fail the import on a missing match.

chapulina (approved):

nuclearsandwich:
The CI failure indicates that ign-math6 is missing a suitable source package for the imported versions. Until that is sorted out we've decided _not_ to update the debian buster config or update the import there at all.

nuclearsandwich:
Due to problems with the way the "Get changed files" action behaved I ended up rebasing out all changes to the Debian Buster configuration.

PR #150, opened 2022-04-15
Update Citadel imports for Ubuntu Focal.

nuclearsandwich:
Uses the script from #144 

[comparison url using the commit after reformatting](https://github.com/ros-infrastructure/reprepro-updater/compare/55f6aaffef2f0cb1379e6320eabe4b11ecca87a8..nuclearsandwich/citadel-update-2022-04-14)

### DIFF
--- a/config/ignition_citadel_gazebo11_ubuntu_focal.yaml
+++ b/config/ignition_citadel_gazebo11_ubuntu_focal.yaml
@@ -116,6 +116,9 @@ filter_formula: "\
  Package (= libignition-sensors3-thermal-camera) |\
  Package (= libignition-sensors3-thermal-camera-dev) \
 ), $Version (% 3.3.0-1*) |\
+(Package (= ignition-tools) |\
+ Package (= libignition-tools-dev) \
+), $Version (% 1.4.1-1*) |\
 (Package (= ignition-transport8) |\
  Package (= libignition-transport8) |\
  Package (= libignition-transport8-core-dev) |\
--- a/config/ignition_citadel_gazebo11_ubuntu_focal.yaml
+++ b/config/ignition_citadel_gazebo11_ubuntu_focal.yaml
@@ -2,7 +2,7 @@ name: ignition_citadel_gazebo11_ubuntu_focal
 method: http://packages.osrfoundation.org/gazebo/ubuntu-stable
 suites: [focal]
 component: main
-architectures: [amd64, i386, armhf, arm64, source]
+architectures: [amd64, armhf, arm64, source]
 filter_formula: "\
 (Package (= gazebo11) |\
  Package (= gazebo11-common) |\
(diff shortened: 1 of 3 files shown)


nuclearsandwich:
The ignition-math6 import is in the same situation as #145. Which I think means we just revert all of the buster changes and proceed with just Ubuntu Focal.

config/ignition_citadel_gazebo11_ubuntu_focal.yaml
```diff
@@ -2,136 +2,144 @@ name: ignition_citadel_gazebo11_ubuntu_focal
...
+ Package (= libignition-common4-av-dev) |\
+ Package (= libignition-common4-core-dev) |\
+ Package (= libignition-common4-dev) |\
+ Package (= libignition-common4-events) |\
+ Package (= libignition-common4-events-dev) |\
+ Package (= libignition-common4-graphics) |\
+ Package (= libignition-common4-graphics-dev) |\
+ Package (= libignition-common4-profiler) |\
+ Package (= libignition-common4-profiler-dev) \
+), $Version (% 4.5.0-1*) |\
```
chapulina, line 32: Citadel uses `common3`. I haven't looked at the script to see where you're getting the versions from, but [gazebodistro](https://github.com/ignition-tooling/gazebodistro/blob/master/collection-citadel.yaml) is a good resource

config/ignition_citadel_gazebo11_ubuntu_focal.yaml
```diff
@@ -2,136 +2,144 @@ name: ignition_citadel_gazebo11_ubuntu_focal
...
+(Package (= ignition-fuel-tools6) |\
+ Package (= libignition-fuel-tools6) |\
+ Package (= libignition-fuel-tools6-dev) \
+), $Version (% 6.2.0-1*) |\
```
chapulina, line 36: It should be `fuel-tools4`

chapulina (changes requested):
humm it looks like this is using Edifice versions, not Citadel

config/ignition_citadel_gazebo11_ubuntu_focal.yaml
nuclearsandwich: Yeah, I lost track of which collections I had actually vetted and which were just copy-paste jobs. I've updated the script to use the versions described in that project (although it's still hard-coded so changes in the gazebodistro repo won't be picked up automatically)

config/ignition_citadel_gazebo11_ubuntu_focal.yaml
nuclearsandwich: Resolved. See https://github.com/ros-infrastructure/reprepro-updater/pull/150#discussion_r851436513

config/ignition_citadel_gazebo11_ubuntu_focal.yaml
```diff
@@ -2,136 +2,140 @@ name: ignition_citadel_gazebo11_ubuntu_focal
...
-Package (% =libignition-sensors3-rendering) |\
-Package (% =libignition-sensors3-rendering-dev) |\
-Package (% =libignition-sensors3-rgbd-camera) |\
-Package (% =libignition-sensors3-rgbd-camera-dev) |\
-Package (% =libignition-sensors3-thermal-camera) |\
-Package (% =libignition-sensors3-thermal-camera-dev)), \
-$Version (% 3.3.0-1~*)) |\
-((Package (% =ignition-tools) |\
-Package (% =libignition-tools-dev)), \
-$Version (% 1.4.1-1~*)) |\
```
chapulina, line 117: `ignition-tools` is now missing

chapulina (approved):
LGTM once `ignition-tools` 1.4.1 is added back

chapulina (approved):

PR #151, opened 2022-04-15
Implement my own file change detection.

nuclearsandwich:
I kept hitting https://github.com/jitterbit/get-changed-files/issues/42 when working on recent PRs. So I decided to see what it would take to get my own approach in.

The answer, it turned out was not much in terms of trickery but a fair bit of tweaking to account for the GitHub actions environment. I am in favor of this "just use git" but it does mean fetching a deep clone rather than a shallow one. For this project that shouldn't be too punishing though.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -80,7 +80,7 @@ jobs:
         id: files
         if: ${{ matrix.job_type == 'CONFIG_TESTER' }}
         run: |
-          git diff --name-only ${{ github.base_ref }}..HEAD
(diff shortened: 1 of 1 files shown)


j-rivero:
Changes looks reasonable to me. Is there a any test run on top of these changes to check if it runs fine?

nuclearsandwich:
> Changes looks reasonable to me. Is there a any test run on top of these changes to check if it runs fine?

This has been merged into https://github.com/ros-infrastructure/reprepro-updater/pull/150 (which was breaking under the existing change detection) and is being used for that PR.

PR #152, opened 2022-04-15
Update Fortress packages in Ubuntu Focal

nuclearsandwich:
Generated from #144.

### DIFF
--- a/config/ignition_fortress_ubuntu_focal.yaml
+++ b/config/ignition_fortress_ubuntu_focal.yaml
@@ -2,48 +2,73 @@ name: ignition_fortress_ubuntu_focal
 method: http://packages.osrfoundation.org/gazebo/ubuntu-stable
 suites: [focal]
 component: main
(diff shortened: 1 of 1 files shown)


chapulina (approved):
All versions LGTM, thanks!

chapulina:
Oh wait, just saw this is Focal. Should we remove them after we have Jammy?

nuclearsandwich:
> Oh wait, just saw this is Focal. Should we remove them after we have Jammy?

Because Fortress is for Rolling/Humble?
Since we've ended up with Focal as a tier 3 platform for Humble, I think we can continue importing Fortress on Focal as long as it's supported by the Gazebo team.

chapulina:
> Since we've ended up with Focal as a tier 3 platform for Humble

Oh I missed that. Yeah makes sense then 👍🏽

nuclearsandwich:
Import jobs:
* https://build.ros.org/job/import_upstream/227/
* https://build.ros2.org/job/import_upstream/388/

PR #154, opened 2022-04-20
Drop xenial and add jammy from import suites.

nuclearsandwich:
We noticed while debugging other issues that the Jammy repositories actually still include colcon-core 0.6.1.


j-rivero (approved):

cottsay (approved):

wjwwood (approved):

nuclearsandwich:
Import jobs:
* https://build.ros.org/job/import_upstream/228/
* https://build.ros2.org/job/import_upstream/389/

PR #156, opened 2022-04-29
Add colcon-hardware-acceleration

cottsay:
Imported to: https://packagecloud.io/dirk-thomas/colcon


nuclearsandwich:
Import upstream jobs:
* https://build.ros.org/job/import_upstream/230/
* https://build.ros2.org/job/import_upstream/391/

PR #160, opened 2022-06-11
Update Fortress config for Ubuntu Jammy.

nuclearsandwich:
Using the current generator script from #144.

The primary objective of this update, although it is not reflected in the config. Is to pull in arm64 packages for Fortress now that they are available. I'll post the bootstrap snapshot diff after import to verify that they were pulled in before running import jobs.


chapulina (approved):

j-rivero (approved):

j-rivero:
ROS2 import job https://build.ros2.org/job/import_upstream/409/

PR #161, opened 2022-06-23
Update to colcon-core 0.9.0

cottsay:
Upstream import is complete: https://packagecloud.io/dirk-thomas/colcon


nuclearsandwich (approved):
All good once the source package checks pass. (And no reason to expect they won't.)

cottsay:
Import jobs:
* [Build Status](https://build.ros.org/job/import_upstream/241/)
* [Build Status](https://build.ros2.org/job/import_upstream/402/)

PR #163, opened 2022-07-25
Rename README to README.md

willie5588912:
Hi, I just rename README to README.md for better format showing on Github, thanks.


cottsay:
I don't think this is markdown, though some elements certainly look like it. Can we change the section blocks to make this a proper markdown file?

PR #165, opened 2022-08-16
Update fortress jammy: ign-cmake, rendering and gazebo6

j-rivero:
Probably fixes https://github.com/gazebosim/gz-cmake/issues/304

### DIFF
--- a/config/ignition_fortress_ubuntu_jammy.yaml
+++ b/config/ignition_fortress_ubuntu_jammy.yaml
@@ -8,7 +8,7 @@ filter_formula: "\
 ), $Version (% 1.0.3-1*) |\
 (Package (= ignition-cmake2) |\
  Package (= libignition-cmake2-dev) \
-), $Version (% 2.12.1-2*) |\
+), $Version (% 2.14.0-2*) |\
 (Package (= ignition-common4) |\
  Package (= libignition-common4) |\
  Package (= libignition-common4-av) |\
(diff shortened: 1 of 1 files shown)


chapulina:
Should we also update Focal? We actually have it duplicate in 2 places, we should remove one of them so we don't need to keep them in sync:

https://github.com/ros-infrastructure/reprepro-updater/blob/34039da99c40192097048bb24247a8240468518c/config/ignition_fortress_ubuntu_focal.yaml#L9-L11

https://github.com/ros-infrastructure/reprepro-updater/blob/34039da99c40192097048bb24247a8240468518c/config/ignition_citadel_gazebo11_ubuntu_focal.yaml#L17-L19

chapulina (approved):
Approving to unblock the fix. We should only need to upgrade Jammy for the fix, since the issue was introduced only on Jammy here:

* https://github.com/ros-infrastructure/reprepro-updater/pull/160

j-rivero:
> Approving to unblock the fix. We should only need to upgrade Jammy for the fix, since the issue was introduced only on Jammy here:

Yep, let's fix jammy by now. We'll update focal later. I've added a couple more of imports.

vatanaksoytezer:
@j-rivero @chapulina thank you for addressing this super quickly!

j-rivero:
I'm pretty sure that this will fix problems at least up to RVIZ deps. Merging and syncing.

PR #168, opened 2022-08-17
Import Citadel gz packages

chapulina:
* Part of https://github.com/gazebo-tooling/release-tools/issues/788

Summary:

* Add all `gz` counterparts to `ignition` packages
* Bump to the latest releases. ⚠️ I've included some versions that are staged to be released, but not there yet:
    - [x] https://github.com/gazebosim/gz-gui/pull/474
    - [x] https://github.com/gazebosim/gz-sim/pull/1657
    - [x] `gz-rendering` 3.6.0-5
* Also removed libraries from Fortress that are duplicates of Citadel, only for Focal. I think it makes sense to have a single source of truth.
* I didn't update Buster on purpose, because that's EOL and we haven't been making new releases for it

Here's the current status of Citadel source packages on packages.osrf (this doesn't check that the `gz` packages exist, I've done that manually 🙃 ):

image

### DIFF
--- a/config/ignition_citadel_gazebo11_ubuntu_focal.yaml
+++ b/config/ignition_citadel_gazebo11_ubuntu_focal.yaml
@@ -12,12 +12,25 @@ filter_formula: "\
  Package (= libgazebo11) |\
  Package (= libgazebo11-dev) \
 ), $Version (% 11.10.2-1*) |\
-(Package (= ignition-citadel) \
-), $Version (% 1.0.2-1*) |\
+(Package (= gz-citadel) \
+ Package (= ignition-citadel) |\
+), $Version (% 1.0.2-2*) |\
 (Package (= ignition-cmake2) |\
+ Package (= libgz-cmake2-dev) \
  Package (= libignition-cmake2-dev) \
-), $Version (% 2.12.1-1*) |\
+), $Version (% 2.14.0-2*) |\
 (Package (= ignition-common3) |\
+ Package (= libgz-common3) |\
+ Package (= libgz-common3-av) |\
+ Package (= libgz-common3-av-dev) |\
+ Package (= libgz-common3-core-dev) |\
+ Package (= libgz-common3-dev) |\
+ Package (= libgz-common3-events) |\
+ Package (= libgz-common3-events-dev) |\
+ Package (= libgz-common3-graphics) |\
+ Package (= libgz-common3-graphics-dev) |\
+ Package (= libgz-common3-profiler) |\
+ Package (= libgz-common3-profiler-dev) \
  Package (= libignition-common3) |\
  Package (= libignition-common3-av) |\
  Package (= libignition-common3-av-dev) |\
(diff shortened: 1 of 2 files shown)


nuclearsandwich:
> * Also removed libraries from Fortress that are duplicates of Citadel, only for Focal. I think it makes sense to have a single source of truth.

I agree that single-source of truth is an important objective.
The reason that there is duplication is that the unit-of-change for these configs is the individual config file so when updating a citadel-on-focal or fortress-on-focal if we want to make sure that each package is it is on the OSRF repos matches the import we need to break sole-source-of-truth to satisfy atomic unit-of-change. And I judge the latter to be more important.

Another possibility is to refactor _all_ gazebo suites into a single config file for each (OS,version) combination so that packages shared among them are not duplicated but this would mean that we'd have to stabilize the OSRF repos an (OS,version) at a time rather than (OS,version,gazebo suite).

config/ignition_citadel_gazebo11_ubuntu_focal.yaml
```diff
@@ -12,12 +12,25 @@ filter_formula: "\
  Package (= libgazebo11) |\
  Package (= libgazebo11-dev) \
 ), $Version (% 11.10.2-1*) |\
```
nuclearsandwich, line 14: According to the generator script there is a new version of gazebo11 to import

```suggestion
), $Version (% 11.11.0-1*) |\
```

config/ignition_citadel_gazebo11_ubuntu_focal.yaml
```diff
@@ -29,38 +42,67 @@ filter_formula: "\
...
+ Package (= libgz-launch2-dev) |\
  Package (= libignition-launch2) |\
  Package (= libignition-launch2-dev) \
-), $Version (% 2.2.2-1*) |\
+), $Version (% 2.3.0-1*) |\
 (Package (= ignition-math6) |\
+ Package (= libgz-math6) |\
+ Package (= libgz-math6-dbg) |\
+ Package (= libgz-math6-dev) |\
+ Package (= libgz-math6-eigen3-dev) |\
```
nuclearsandwich, line 78:
```suggestion
 Package (= libgz-math6-eigen-dev) |\
```

According to the introspection performed by the generator script there is no package named `libgz-math6-eigen3-dev` but the `libignition-math6-eigen3-dev` package does use `eigen3` instead of `eigen`. @j-rivero do you know if this is a typo / omission or if the plan is to rename the libgz... package?

config/ignition_fortress_ubuntu_focal.yaml
```diff
@@ -6,9 +6,6 @@ architectures: [amd64, armhf, arm64, source]
 filter_formula: "\
 (Package (= ignition-fortress) \
 ), $Version (% 1.0.3-1*) |\
-(Package (= ignition-cmake2) |\
- Package (= libignition-cmake2-dev) \
-), $Version (% 2.12.1-1*) |\
```
nuclearsandwich, line 11: I would prefer that these changes are revered for the reasons described in https://github.com/ros-infrastructure/reprepro-updater/pull/168#issuecomment-1218663623

config/ignition_fortress_ubuntu_focal.yaml
```diff
@@ -41,14 +38,6 @@ filter_formula: "\
...
  Package (= libignition-launch5-dev) \
 ), $Version (% 5.1.0-1*) |\
-(Package (= ignition-math6) |\
- Package (= libignition-math6) |\
- Package (= libignition-math6-dbg) |\
- Package (= libignition-math6-dev) |\
- Package (= libignition-math6-eigen3-dev) |\
- Package (= python3-ignition-math6) |\
- Package (= ruby-ignition-math6) \
-), $Version (% 6.10.0-1*) |\
```
nuclearsandwich, line 51: I would prefer that these changes are revered for the reasons described in https://github.com/ros-infrastructure/reprepro-updater/pull/168#issuecomment-1218663623

config/ignition_fortress_ubuntu_focal.yaml
```diff
@@ -69,11 +58,6 @@ filter_formula: "\
  Package (= libignition-physics5-tpelib) |\
  Package (= libignition-physics5-tpelib-dev) \
 ), $Version (% 5.1.0-2*) |\
-(Package (= ignition-plugin) |\
- Package (= libignition-plugin) |\
- Package (= libignition-plugin-dbg) |\
- Package (= libignition-plugin-dev) \
-), $Version (% 1.2.1-1*) |\
```
nuclearsandwich, line 76: I would prefer that these changes are revered for the reasons described in https://github.com/ros-infrastructure/reprepro-updater/pull/168#issuecomment-1218663623

config/ignition_fortress_ubuntu_focal.yaml
```diff
@@ -118,9 +102,6 @@ filter_formula: "\
  Package (= libignition-sensors6-thermal-camera) |\
  Package (= libignition-sensors6-thermal-camera-dev) \
 ), $Version (% 6.3.0-2*) |\
-(Package (= ignition-tools) |\
- Package (= libignition-tools-dev) \
-), $Version (% 1.4.1-1*) |\
```
nuclearsandwich, line 123: I would prefer that these changes are revered for the reasons described in https://github.com/ros-infrastructure/reprepro-updater/pull/168#issuecomment-1218663623

nuclearsandwich (changes requested):
Overall this is very close to what the generator script produces. I caught a couple of discrepancies which need to be resolved and I'm going to recommend not changing the fortress config although I understand the motivation behind it I think there is an overriding reason to leave it as is (described in thread).

config/ignition_citadel_gazebo11_ubuntu_focal.yaml
chapulina: I thought I had fixed that

https://github.com/gazebo-release/gz-math6-release/commit/3b8555972bd25344edbc3501cd9cec11a96f43d9

j-rivero: yep, eigen3 should be used.

config/ignition_fortress_ubuntu_focal.yaml
j-rivero: Reverted [adf0dce](https://github.com/ros-infrastructure/reprepro-updater/pull/168/commits/adf0dcee76545baaff671047521d91dad91e8176)

config/ignition_fortress_ubuntu_focal.yaml
j-rivero: Reverted [adf0dce](https://github.com/ros-infrastructure/reprepro-updater/pull/168/commits/adf0dcee76545baaff671047521d91dad91e8176)

config/ignition_fortress_ubuntu_focal.yaml
j-rivero: Reverted [adf0dce](https://github.com/ros-infrastructure/reprepro-updater/pull/168/commits/adf0dcee76545baaff671047521d91dad91e8176)

config/ignition_fortress_ubuntu_focal.yaml
j-rivero: Reverted [adf0dce](https://github.com/ros-infrastructure/reprepro-updater/pull/168/commits/adf0dcee76545baaff671047521d91dad91e8176)

config/ignition_citadel_gazebo11_ubuntu_focal.yaml
nuclearsandwich: The generator script is still finding the bare eigen package name on Focal. Looking at the package pool it appears that only jammy has the corrected package name.
eigen

nuclearsandwich (changes requested):
The eigen / eigen3 issue appears to've been fixed for jammy only so the focal package still lists plain eigen. Once that is corrected this is good to go (assuming no other releases happen in the meantime).

j-rivero:
> The eigen / eigen3 issue appears to've been fixed for jammy only so the focal package still lists plain eigen. Once that is corrected this is good to go (assuming no other releases happen in the meantime).

Yup, they are buggy right now. Fix in https://github.com/gazebo-release/gz-math6-release/pull/17. I'll release a -3 and update this PR. Thanks for checking.

nuclearsandwich:
I didn't know that magic comments could close PRs across organization boundaries. Reopened after it was (presumably unintentionally) closed by a comment in https://github.com/gazebo-release/gz-math6-release/pull/17

j-rivero:
> I didn't know that magic comments could close PRs across organization boundaries. Reopened after it was (presumably unintentionally) closed by a comment in [gazebo-release/gz-math6-release#17](https://github.com/gazebo-release/gz-math6-release/pull/17)

Ouch, totally unintentionally yes.

j-rivero:
The data matches the output of the script in #144

j-rivero:
After import to bootstrap, I've checked the status of that repo in https://build.osrfoundation.org/job/ignition_citadel-install-pkg_ros_bootstrap-any-manual/2/ . Code coming from https://github.com/gazebo-tooling/release-tools/pull/801. Seems fine to my eyes.

j-rivero:
Imported https://build.ros2.org/job/import_upstream/417/

nuclearsandwich:
> Imported https://build.ros2.org/job/import_upstream/417/

In general please run import jobs for build.ros.org and build.ros2.org together so that future build.ros.org import jobs don't have excess changes.

PR #169, opened 2022-08-17
Import Fortress gz packages

chapulina:
* Builds on top of https://github.com/ros-infrastructure/reprepro-updater/pull/168
* Part of https://github.com/gazebo-tooling/release-tools/issues/788

Summary:

* Add all `gz` counterparts to `ignition` packages
* Bump to the latest releases. ⚠️ I've included some versions that are staged to be released, but not there yet:
    - [x] `gz-sim` 6.11.0-2
    - [x] `gz-utils` 1.4.0-2
* I didn't update Buster on purpose, because that's EOL and we haven't been making new releases for it

Here's the current status of Fortress source packages on packages.osrf (this doesn't check that the `gz` packages exist, I've done that manually 🙃 ):

image

### DIFF
--- a/config/ignition_fortress_ubuntu_focal.yaml
+++ b/config/ignition_fortress_ubuntu_focal.yaml
@@ -116,10 +116,13 @@ filter_formula: "\
  Package (= libignition-physics5-tpelib-dev) \
 ), $Version (% 5.2.0-2*) |\
 (Package (= ignition-plugin) |\
+ Package (= libgz-plugin) |\
+ Package (= libgz-plugin-dbg) |\
+ Package (= libgz-plugin-dev) |\
  Package (= libignition-plugin) |\
  Package (= libignition-plugin-dbg) |\
  Package (= libignition-plugin-dev) \
-), $Version (% 1.2.1-1*) |\
+), $Version (% 1.3.0-1*) |\
 (Package (= ignition-rendering6) |\
  Package (= libgz-rendering6) |\
  Package (= libgz-rendering6-core-dev) |\
--- a/config/ignition_fortress_ubuntu_focal.yaml
+++ b/config/ignition_fortress_ubuntu_focal.yaml
@@ -50,7 +50,7 @@ filter_formula: "\
  Package (= libignition-gazebo6-dbg) |\
  Package (= libignition-gazebo6-dev) |\
  Package (= libignition-gazebo6-plugins) |\
- Package (= python3-gz-sim6) \
+ Package (= python3-gz-sim6) |\
  Package (= python3-ignition-gazebo6) \
 ), $Version (% 6.11.0-2*) |\
 (Package (= ignition-gui6) |\
--- a/config/ignition_fortress_ubuntu_focal.yaml
+++ b/config/ignition_fortress_ubuntu_focal.yaml
@@ -233,10 +233,6 @@ filter_formula: "\
  Package (= libignition-transport11-log-dev) \
 ), $Version (% 11.2.0-3*) |\
 (Package (= ignition-utils1) |\
- Package (= libgz-utils1) |\
- Package (= libgz-utils1-cli-dev) |\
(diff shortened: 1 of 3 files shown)


nuclearsandwich:
> (this doesn't check that the `gz` packages exist, I've done that manuall

According to the generator script libgz-transport11-cli is still outstanding:
@j-rivero do you know which debbuilder job is responsible for this package?

```
libgz-transport11-cli 11.2.0-2~focal in gz-transport11-cli does not match expected version 11.2.0-3~focal
```

j-rivero:
> According to the generator script libgz-transport11-cli is still outstanding:
> @j-rivero do you know which debbuilder job is responsible for this package?

`libgz-transport11-cli` was an error and a wrong package produced by the error that really should have been `gz-transport11-cli`. Should not be imported.

config/ignition_fortress_ubuntu_focal.yaml
```diff
@@ -117,24 +210,38 @@ filter_formula: "\
...
  Package (= libignition-transport11-dev) |\
  Package (= libignition-transport11-log) |\
  Package (= libignition-transport11-log-dev) \
-), $Version (% 11.0.0-1*) |\
+), $Version (% 11.2.0-3*) |\
 (Package (= ignition-utils1) |\
+ Package (= libgz-utils1) |\
+ Package (= libgz-utils1-cli-dev) |\
+ Package (= libgz-utils1-dbg) |\
+ Package (= libgz-utils1-dev) |\
```
nuclearsandwich, line 239: According to a fresh run of the generator script, these packages don't yet exist on Focal.
I don't object to assuming they will be eventually but I also see a benefit in waiting for them to appear so that the generator script will help catch any discrepancies between expectation and reality. The import process does not error if an individual package in the filter fails to find a match, which is one of the reasons the generator removes packages from the config which aren't found in the repositories. Between the two processes, we guarantee that exactly the packages available are imported and that each of them has a consistent version and corresponding source package.

In the course of writing all that out, I actually talked myself into removing these so that the generator output matches exactly. That way if it's an "error" that these packages are absent we catch it now. I'm pushing a change now to that effect.

nuclearsandwich (approved):
With the caveat that I've just pushed a bunch of my own changes that will be good for @j-rivero to look over. I think this is now ready to go _once the Citadel PR (#168) it's based on is merged_.

Most of my changes were specifically related to updating to the latest version available in the gazebo repositories which is absolutely necessary for a successful complete import and including addition gz packages which have been added. In one case, I removed listed gz packages which are not present in the upstream repository.

config/ignition_fortress_ubuntu_focal.yaml
j-rivero: We have releasing them, they were missing.

j-rivero:
CI is happy, script in #144 is happy. Merging.

j-rivero:
Tested different combinations using https://github.com/gazebo-tooling/release-tools/pull/801:
 * Jammy amd64 https://build.osrfoundation.org/job/ignition_fortress-install-pkg_ros_bootstrap-any-manual/5/
 * Jammy arm64 https://build.osrfoundation.org/job/ignition_fortress-install-pkg_ros_bootstrap-any-manual/11/
 * Focal amd64 https://build.osrfoundation.org/job/ignition_fortress-install-pkg_ros_bootstrap-any-manual/6/
 * Focal arm64 https://build.osrfoundation.org/job/ignition_fortress-install-pkg_ros_bootstrap-any-manual/12/

j-rivero:
Imported https://build.ros2.org/job/import_upstream/419/

nuclearsandwich:
> Imported https://build.ros2.org/job/import_upstream/419/

In the future please run import_upstream for both build farms regardless of which one is "targeted" by a chance in order to reduce noise in other import jobs related to these changes.

PR #171, opened 2022-09-02
Update ubuntu focal config for fortress.

nuclearsandwich:
Trying to wrap up https://github.com/ros/rosdistro/pull/34177.

This import should resolve the current failure to find `libgz-utils1-dev`.

### DIFF
--- a/config/ignition_fortress_ubuntu_focal.yaml
+++ b/config/ignition_fortress_ubuntu_focal.yaml
@@ -10,7 +10,7 @@ filter_formula: "\
 (Package (= ignition-cmake2) |\
  Package (= libgz-cmake2-dev) |\
  Package (= libignition-cmake2-dev) \
-), $Version (% 2.14.0-2*) |\
+), $Version (% 2.15.0-1*) |\
 (Package (= ignition-common4) |\
  Package (= libgz-common4) |\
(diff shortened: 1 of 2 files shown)


config/ignition_fortress_ubuntu_focal.yaml
```diff
@@ -10,7 +10,7 @@ filter_formula: "\
 (Package (= ignition-cmake2) |\
  Package (= libgz-cmake2-dev) |\
  Package (= libignition-cmake2-dev) \
-), $Version (% 2.14.0-2*) |\
+), $Version (% 2.15.0-1*) |\
```
chapulina, line 13: Can we also change it for Citadel to keep them in sync?

https://github.com/ros-infrastructure/reprepro-updater/blob/86c07162a69626445dd06430c2a701ece3985105/config/ignition_citadel_gazebo11_ubuntu_focal.yaml#L21

chapulina (approved):

config/ignition_fortress_ubuntu_focal.yaml
nuclearsandwich: Good call. I ran a full update of citadel on focal as well.

In retrospect we ought to have organized these configs around debian/ubuntu version rather than gazebo suite.

PR #172, opened 2022-09-29
Add colcon-meson package to colcon imports.

nuclearsandwich:
Assuming that everything worked colcon-meson 0.4.2 is now available in the colcon apt repositories for us to import.

The package only currently supports Ubuntu Jammy due to the meson >= 0.6 requirement.

### DIFF
--- a/config/colcon.debian.upstream.yaml
+++ b/config/colcon.debian.upstream.yaml
@@ -20,6 +20,7 @@ filter_formula: "\
 (Package (% python3-colcon-installed-package-information), $Version (% 0.1.0*))|\
 (Package (% python3-colcon-lcov-result), $Version (% 0.5.0*))|\
 (Package (% python3-colcon-library-path), $Version (% 0.2.1*))|\
+(Package (% python3-colcon-meson), $Version (% 0.4.2*))|\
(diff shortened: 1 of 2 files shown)


config/colcon.debian.upstream.yaml
```diff
@@ -20,6 +20,7 @@ filter_formula: "\
 (Package (% python3-colcon-installed-package-information), $Version (% 0.1.0*))|\
 (Package (% python3-colcon-lcov-result), $Version (% 0.5.0*))|\
 (Package (% python3-colcon-library-path), $Version (% 0.2.1*))|\
+(Package (% python3-colcon-meson), $Version (% 0.4.2*))|\
```
nuclearsandwich, line 23: Even though there is not technically a Debian release which includes this package I think it makes sense to keep the debian and ubuntu files in sync.

nuclearsandwich:
FYI @christianrauch.

cottsay (approved):

nuclearsandwich:
Import jobs
* https://build.ros.org/job/import_upstream/251/
* https://build.ros2.org/job/import_upstream/422/

PR #173, opened 2022-10-14
Handle removefilter args that exceed command line max length.

nuclearsandwich:
This is a kick-off PR since I am flying today. There are plenty of valid ways to break this one down. I was planning to try 1 partition, then 2, then 4, and so on. One challenge is that package names are not all the same length so it is possible that in two partitions the first may be small enough but the second would not so all partitions have to be templated as commands and checked before running any.

Since we already have a fully enumerated set of packages to invalidate I don't think there is any ordering requirement of which packages we remove in which order so long as the overall operation only reports success if all commands succeed.

### DIFF
--- a/src/reprepro_updater/helpers.py
+++ b/src/reprepro_updater/helpers.py
@@ -8,8 +8,6 @@ import time
 
 from reprepro_updater.repository_info import RepositoryInfo
 
-ARG_MAX_LEN = os.sysconf('SC_ARG_MAX')
-
 
 class LockContext:
     def __init__(self, lockfilename=None, timeout=3000):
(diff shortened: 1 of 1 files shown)


src/reprepro_updater/helpers.py
```diff
@@ -92,10 +94,13 @@ def invalidate_packages(repo_dir, distro, arch, packages):
...
     filterstr = 'Package (== {})'
     filterlist = [filterstr.format(pkgname) for pkgname in packages]
-    invalidate_packages_command = ['reprepro', '-b', repo_dir,
-                                   '-T', 'deb', '-A', arch, '-V',
-                                   'removefilter', distro,
-                                   ' | '.join(filterlist)]
+    cmd_prefix = ['reprepro', '-b', repo_dir, '-T', 'deb',
+                  '-A', arch, '-V', 'removefilter', distro]
+    invalidate_packages_command = cmd_prefix + [' | '.join(filterlist)]
+    if len(invalidate_packages_command) > ARG_MAX_LEN:
```
cottsay, line 100: Won't this just give you the length of the array?

nuclearsandwich: Yes it will. Shows what comes of rushing. I can't recall if SC_ARG_MAX applies to an individual arg or the entire line.

I'll visit the man pages and revise this.

clalancette:
FYI, I went ahead and made some changes here to try to make things work better.

In particular, what I did was to first attempt to remove the entire list in one `reprepro` command.  If that succeeds, then we are done.  If that fails, I then split the list in half and try again.  This won't fix the problem for all time, but it should at least double the amount of packages we can support with this.

The main reason I went with a solution like this is that I couldn't reconcile the value of SC_ARG_MAX with what was failing.  In particular, I was getting a value of 2097152 for SC_ARG_MAX, but the failing command is only of length 131072.  Some other limit must be kicking in, but I don't know what it is.

Please take a look and let me know what you think.

cottsay (approved):

j-rivero (approved):

nuclearsandwich:
> The main reason I went with a solution like this is that I couldn't reconcile the value of SC_ARG_MAX with what was failing. In particular, I was getting a value of 2097152 for SC_ARG_MAX, but the failing command is only of length 131072. Some other limit must be kicking in, but I don't know what it is.

I hadn't been able to validate it on the plane but I think that the full environment may be included in what counts against SC_ARG_MAX. It's also possible that we're hitting an earlier limit than the one specified by sysconf.

Empirically trying and retrying is a good solution. I think an additional step here would be to generalize the partitioning so that if a half-size list is _still_ too long it splits into fourths and so on but that gets harder as it won't necessarily be the first attempted split which fails as package names have differing lengths.

Practically speaking, the change as is will probably be sufficient indefinitely.

PR #176, opened 2022-12-12
Add colcon-clean package

cottsay:
Imported here: https://packagecloud.io/dirk-thomas/colcon

This package only targets Jammy and Bullseye due to dependency availability.


cottsay:
[Build Status](https://build.ros2.org/job/import_upstream/427/)
[Build Status](https://build.ros.org/job/import_upstream/256/)

PR #178, opened 2023-02-02
Update a handful of colcon package versions

cottsay:
* colcon-core 0.12.0
* colcon-defaults 0.2.8
* colcon-mixin 0.2.3
* colcon-python-setup-py 0.2.8

Imported here: https://packagecloud.io/dirk-thomas/colcon

I'll note that `colcon-python-setup-py` was released a while ago but it never got imported here.


nuclearsandwich (approved):

cottsay:
[Build Status](https://build.ros.org/job/import_upstream/258/)
[Build Status](https://build.ros2.org/job/import_upstream/429/)

PR #181, opened 2023-07-07
Remove -dbgsym packages from diff by default

peci1:


ros-discourse:
This pull request has been mentioned on **ROS Discourse**. There might be relevant details there:

https://discourse.ros.org/t/new-packages-for-iron-irwini-2023-06-30/32169/6

scripts/diff_packages.py
```diff
@@ -46,6 +46,7 @@ def main():
     parser.add_argument('tofile', type=str, nargs='+', default=None)
     parser.add_argument('rosdistro', type=str, default='groovy')
     parser.add_argument('--output-dir', dest='output_dir', type=str, default='.')
+    parser.add_argument('--with-debug-pkgs', action='store_true')
```
tfoote, line 49: I don't think that this is best way to go. The maintainer/reporter doesn't get a second chance to run this as it's run as part of the sync process. The maintainer just pulls the results from the logs. And we need the buildfarm logs to be complete. And to that end I don't think that we want to ever actually suppress this part of the report. If we were to use this we'd have to build the reports for every platform twice. I think that it might make sense to segregate the report such that it's easy for the maintainer to copy and paste the results but skip the debug packages by separating them out slightly into a separate section or something like that. 

Even better would be to actually collapse the line and add an addendum which indicates that the dbgsym was copied too as long as it matched the package naming and version in the main list. 

```
 * [ros-iron-mrpt2](https://www.mrpt.org/): 2.9.0-2 -> 2.9.3-1
 * ros-iron-mrpt2-dbgsym: 2.9.0-2 -> 2.9.3-1
 ```
 
```
 * [ros-iron-mrpt2](https://www.mrpt.org/): 2.9.0-2 -> 2.9.3-1 (with `-dbgsym`)
```

This would have the significant value that if you see a `-dbgsym` in the main list it clearly indicates that something is out of sync. We could even flag an umatched `dbgsym` with a warning.

**Review by tfoote — CHANGES_REQUESTED** [review]

**Comment by peci1** [comment]
Okay, I haven't known there isn't a way to run this script locally. I updated the PR so that the dbgsym packages are not counted in the totals and are merged to the line with their base package:

```
Added Packages [16]:
 * ros-noetic-chaplus-ros: 2.1.26-1focal.20230620.184640
 * ros-noetic-cras-imu-tools: 1.0.1-1focal.20230620.195124 (+debug symbols ros-noetic-cras-imu-tools-dbgsym)
```

As the debug packages can be (in theory) added/removed independently of their base packages, I tried to reflect their added/updated/removed state by using `+debug symbols`, `debug symbols` and `-debug symbols`. I'm not sure whether this is worth it, though.

PR #189, opened 2023-10-05
Update target platforms for colcon upstream repos

cottsay:
This matches what all future releases of colcon packages are expected to support.

### DIFF
--- a/config/colcon.ubuntu.upstream.yaml
+++ b/config/colcon.ubuntu.upstream.yaml
@@ -1,6 +1,6 @@
 name: colcon
 method: https://packagecloud.io/dirk-thomas/colcon/ubuntu
-suites: [bionic, focal, jammy]
+suites: [focal, jammy]
 component: main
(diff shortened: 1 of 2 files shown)


tfoote (approved):
The synatax looks correct to me as well as the supported versions.

nuclearsandwich (approved):

config/colcon.ubuntu.upstream.yaml
```diff
@@ -1,6 +1,6 @@
 name: colcon
 method: https://packagecloud.io/dirk-thomas/colcon/ubuntu
-suites: [bionic, focal, jammy]
```
nuclearsandwich, line 3: I'm assuming that the previous import successfully picked up the bionic releases that you were making to clean things up.

cottsay: Yes - 0.14.0 is in the ROS public repositories now.

PR #194, opened 2024-01-22
Update to colcon-core 0.15.2

cottsay:
https://packagecloud.io/dirk-thomas/colcon


nuclearsandwich:
When investigating, I thought that the reason for the mismatched file size could be an issue of not having apt-https-transport installed and getting redirect messages interpreted as payloads but that was not the case. The problem does appear to be incorrect metadata in <https:packagecloud.io/dirk-thomas/colcon/debian/dists/bullseye/main/source/Sources>

j-rivero (approved):

cottsay:
[Build Status](https://build.ros2.org/job/import_upstream/454/)
[Build Status](https://build.ros.org/job/import_upstream/279/)

PR #198, opened 2024-02-13
Rework some sharp edges in GitHub Actions

cottsay:
Two high-level changes here:
* Drop the matrix in favor of conditionally running steps
   The job runs so fast that I don't think there's any advantage in parallelizing using the matrix
* Use base ref SHA for determining which files changed
   The intent here is to make PRs from forks work by not referencing `origin`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,21 +9,8 @@ on:
 
 
 jobs:
-  bionic-ci:
+  focal-ci:
     runs-on: ubuntu-20.04
-    name: Focal Aptly ${{ matrix.job_type }} -
-          debug ${{ matrix.show_debug_cmds }}
-    strategy:
-      matrix:
-        show_debug_cmds:
-          - true
-          - false
-        job_type:
-          - 'CI'
-          - 'CONFIG_TESTER'
-        exclude:
-          - job_type: 'CONFIG_TESTER'
-            show_debug_cmds: false
     steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -9,21 +9,8 @@ on:
 
 
 jobs:
-  bionic-ci:
+  focal-ci:
```
nuclearsandwich, line 12: Does it make sense to drop the ubuntu distro in the name and just call this job `ci`?

nuclearsandwich (approved):
Thanks, this is a lot cleaner to read!

.github/workflows/ci.yml
cottsay: I agree, done in cd88e88.

j-rivero:
Worked after I tried it on my fork here https://github.com/j-rivero/reprepro-updater/pull/5

.github/workflows/ci.yml
```diff
@@ -69,21 +56,31 @@ jobs:
 
       # CI TESTING
       - name: Run CI tests
```
j-rivero, line 58:
```suggestion
      - name: Run software testsuite
```
Reading the whole thing today, seems clear to me to specify that what we are running is the testsuite. Here and in the debug step.

j-rivero (approved):
Minor suggestion to rename the CI steps. Looks good.

PR #200, opened 2024-02-19
Add Noble and Trixie imports for colcon

cottsay:


Crola1702 (approved):

cottsay:
I believe that performing a normal import procedure should make the ROS repos "catch up" on the missing imports.

PR #212, opened 2024-09-26
Update to colcon-core 0.18.1

cottsay:
Imported here: https://packagecloud.io/app/dirk-thomas/colcon/search?q=colcon-core_0.18.0&filter=all&filter=all&dist=

And not on a Friday this time! Look at that!


nuclearsandwich (approved):
I think there was some discussion that there are issues with tomli on Focal so you may opt not to merge this, but if you do or want to use this for an update after dropping focal, works for me.

cottsay:
[Build Status](https://build.ros.org/job/import_upstream/302/)
[Build Status](https://build.ros2.org/job/import_upstream/480/)

PR #219, opened 2024-12-02
Import Gazebo11 11.15.0

j-rivero:
Getting the latest planned version of Gazebo11 into the ROS Repositories. //cc @azeey

### DIFF
--- a/config/ignition_citadel_gazebo11_ubuntu_focal.yaml
+++ b/config/ignition_citadel_gazebo11_ubuntu_focal.yaml
@@ -11,7 +11,7 @@ filter_formula: "\
  Package (= gazebo11-plugin-base) |\
  Package (= libgazebo11) |\
  Package (= libgazebo11-dev) \
-), $Version (% 11.11.0-1*) |\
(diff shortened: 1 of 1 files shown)


cottsay (approved):

j-rivero:
In [7094185](https://github.com/ros-infrastructure/reprepro-updater/pull/219/commits/7094185588388b6a9f08fe85968adac4a548cee5) I bumped the ignition libraries that the CI complained about but needs to inspect why is this set that is not really a dependency of Gazebo11 and other libs (like sdformat) do not complain even if they are in the latest version.

cottsay (approved):

j-rivero:
Sorry for all the changes in the open pull request @cottsay . I finally went for bumping all the libraries to the latest versions to avoid possible conflicts with some bumps requiring new bumps in their dependencies and to assure that the final experience between packages.o.o and packages.ros.o is the same.

cottsay (approved):

j-rivero:
Superseded by 220

PR #228, opened 2025-03-14
24.04 migration

Crola1702:
Fixes:
* Deprecated SafeConfigParser usage. Replaced with ConfigParser
* SyntaxWarning of empty string comparison


nuclearsandwich (approved):
Both of these changes are compatible with Python 3.8 so they can be merged without waiting for an update to the repo host.

PR #266, opened 2026-07-27
Test all suites even when --only-mirror-creation

cottsay:
I think this may have been an oversight in the original change. Either way, I think it would be useful to test ALL the suites when this flag is specified and not only the first one listed.

Follows #107


scripts/aptly/aptly_importer.py
j-rivero: Probably the simulate_repo_import has the same problem, exit should be a continue?

cottsay: Excellent catch. Done in f5e171b.

j-rivero (approved):

PR #267, opened 2026-07-27
Fix aptly_importer.py for repos without sources

cottsay:
It doesn't appear that this script ever worked correctly on repo configurations without sources. This change cleans up the removal of 'sources' from the arch list and records its presence to conditionally pass `-with-sources` to the aptly mirror creation later.

At present, there are three configs that fit this criteria:
- namniart_armhf
- pcl_xenial_armhf_backport
- rticonnextdds730

I'd like to add another one for importing GurumDDS.

### DIFF
--- a/scripts/aptly/aptly_importer.py
+++ b/scripts/aptly/aptly_importer.py
@@ -166,9 +166,11 @@ class UpdaterConfiguration():
             self.architectures = self.config['architectures']
             # source was accepted as a valid architecture to indicate that
             # source packages need to be download. It is not a valid arch in aptly
-            if 'source' in self.config['architectures']:
-                self.architectures = self.config['architectures'].remove('source')
-            self.architectures = self.config['architectures']
+            if 'source' in self.architectures:
+                self.architectures.remove('source')
+                self.with_sources = True
+            else:
+                self.with_sources = False
             self.component = self.config['component']
             self.filter_formula = self.reprepro2aptly.convert(
                 self.config['filter_formula'])
@@ -218,9 +220,11 @@ class UpdaterManager():
         if self.aptly.exists(Aptly.ArtifactType.MIRROR, mirror_name):
             self.__log_ok('Removing existing mirror')
             self.aptly.run(['mirror', 'drop', mirror_name])
-        create_cmd = ['mirror', 'create', '-with-sources',
+        create_cmd = ['mirror', 'create',
                       f"-architectures={','.join(self.config.architectures)}",
                       f"-filter={self.config.filter_formula}"]
+        if self.config.with_sources:
+            create_cmd += ['-with-sources']
         if self.ignore_mirror_signature:
             create_cmd += ['-ignore-signatures']
         create_cmd += [mirror_name,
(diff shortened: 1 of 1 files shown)


scripts/aptly/aptly_importer.py
```diff
@@ -218,9 +220,11 @@ def __create_aptly_mirror(self, distribution):
         if self.aptly.exists(Aptly.ArtifactType.MIRROR, mirror_name):
             self.__log_ok('Removing existing mirror')
             self.aptly.run(['mirror', 'drop', mirror_name])
-        create_cmd = ['mirror', 'create', '-with-sources',
+        create_cmd = ['mirror', 'create',
```
j-rivero, line 223: If we are not passing a value to `-with-sources` we are failling back to aplty default or environmental configuration which can be true and then we do nothing on the system. Let's be explicit, something like:

```diff
-        create_cmd = ['mirror', 'create',
-                      f"-architectures={','.join(self.config.architectures)}",
-                      f"-filter={self.config.filter_formula}"]
-        if self.config.with_sources:
-            create_cmd += ['-with-sources']
+        create_cmd = ['mirror', 'create',
+                      f"-architectures={','.join(self.config.architectures)}",
+                      f"-filter={self.config.filter_formula}",
+                      f"-with-sources={'true' if self.config.with_sources else 'false'}"]
```

scripts/aptly/aptly_importer.py
j-rivero, line 320: We probably need here a check for `if self.config.with_sources:` to avoid problems on checking

j-rivero (changes requested):

scripts/aptly/aptly_importer.py
cottsay: Good call. Done in 52d4fe6.

scripts/aptly/aptly_importer.py
cottsay: So I thought so as well, but it appears that it isn't 100% necessary.

At least for the GurumDDS packages, they seem to explicitly indicate that there are no source packages, so this part of the code is used, circumventing the requirement: https://github.com/ros-infrastructure/reprepro-updater/blob/master/scripts/aptly/aptly_importer.py#L101-L103

So maybe we want to leave this in as a "check" to make sure that we didn't accidentally miss a `source` suite that should actually be present?

scripts/aptly/aptly_importer.py
```diff
@@ -218,9 +220,12 @@ def __create_aptly_mirror(self, distribution):
...
+                      f"-with-sources={'true' if self.config.with_sources else 'false'}",
                       f"-architectures={','.join(self.config.architectures)}",
                       f"-filter={self.config.filter_formula}"]
+        if self.config.with_sources:
```
j-rivero: Need to remove these two lines already integrated in the explicit call.

j-rivero (commented):
One more fix and we are ready

scripts/aptly/aptly_importer.py
cottsay: Ah, whoops! Good call. Done in 58adc38.

j-rivero:
good to my eyes

j-rivero (approved):

PR #270, opened 2026-08-06
Update to colcon-core 0.21.1

cottsay:
Imported here: https://packagecloud.io/app/dirk-thomas/colcon/search?q=colcon-core_0.21.1


cottsay:
[Build Status](https://build.ros2.org/job/import_upstream/557/)

